PR #2, opened 2021-05-06
draft of UI to work with

thomascastleman:
Added a `Transformation` component that serves as the UI for configuring a new transformation. You select an input table (just one for now) from a dropdown, select a transformation type (i.e. filter), type in your transformation (i.e. `Age > 35`) then hit transform.

Needs to be connected up with Jason's CODAP phone interface so we can get actual data contexts from CODAP, and also create them when the transformation goes through. 

You can see it in action by running CODAP as well and going to 
```
http://localhost:4020/dg?di=http://localhost:3000
```


P-bibs:
I have a special appreciation for the exclamation mark on the "Transform!" button. Makes me feel like a mad scientist. 

This look good to merge to me. We could also think about deleting some of the boilerplate like `App.tsx` or maybe using `App.tsx` as the toplevel component and have it render `Transform`

stardust66:
LGTM too

PR #3, opened 2021-05-06
Add small expression language

P-bibs:
The language supports the arithmetic operators `+-*/` as well as parentheses for grouping and `==` for equality testing. It should be whitespace insensitive. You can check the test files for more examples. 

It also supports identifiers, although I haven't tested that much since there's no assignment so the only valid identifiers will be those that we prefill the environment with. There's also not much (anything) in terms of exposing helpful error messages.

Examples:
```javascript
1 + 2 * 3 // 7
(1 + 2) * 3 // 9
1 +             5 // 6
```

EDIT: grammar-ish thing:
```
expr := IDENTIFIER
      | NUMBER
      | true
      | false
      | <expr> <binop> <expr>

binop := +
       | -
       | *
       | /
       | ==
       | !=
       | >
       | >=
       | &&
       | and
       | ||
       | or
```

### DIFF
--- a/src/language/__tests__/integration.test.ts
+++ b/src/language/__tests__/integration.test.ts
@@ -35,7 +35,7 @@ test("correctly processes boolean expressions", () => {
 });
 
 test("correctly processes inequality expressions", () => {
-  let source = "1 > 2 and 1 >= 1";
+  let source = "2 > 1 and 1 >= 1";
   expect(interpret(parse(lex(source)))).toStrictEqual({
     kind: "Bool",
     content: true,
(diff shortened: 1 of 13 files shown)


thomascastleman:
Looks good! Though we may want to add a `>`, `>=`, etc as they make good test cases for filters? What do you all think?

Also perhaps some logical operators so we can put together complex conditions for filters. I know we don't need the full language yet but those might be nice as a proof of concept.

P-bibs:
We've now got logical operations with `and`/`&&`/`or`/`||` and more equality operations with `!=`/`>`/`>=`, as well as boolean literals `true` and `false`. Do you think this is enough for a proof of concept @thomascastleman ?

thomascastleman:
Nice, looks good!

thomascastleman (approved):

P-bibs:
@stardust66 any thoughts on this?

stardust66:
Looks good to me too!

stardust66 (approved):

PR #4, opened 2021-05-07
WIP: Getting started on UI and codapPhone integration

stardust66:
Not for merging. Putting this here in case this is helpful as a start.


src/Transformation.tsx
```diff
@@ -73,18 +92,23 @@ function Transformation() {
...
 
-    // TODO: Do transformation here
-
-    const newContext = await(createDataset("Testing", data));
+    const newData = data.filter(dataItem => {
+      const dataEnv = dataItemToEnv(dataItem);
+      const result = evaluate(transformPgrm, dataEnv);
+      if (result.kind !== "Bool") {
+        setErrMsg("Expected boolean output, instead got number.");
+        return false;
```
stardust66, line 106: It would be nicer to abort here. I can only think of throwing an error and using `try ... catch`. Is there a better option?

thomascastleman: `try ... catch` seems fine to me, we can also wrap the `evaluate` so interpreter errors get caught as well.

P-bibs: Could we turn the filter into an equivalent for loop and then use return to bail out early? I agree about try/catch for evaluate though, the parser/lever/interpreter all throw errors that could be surfaced to help us fix typos in formulas

thomascastleman (commented):
@stardust66 @P-bibs what do you all think of this implementation of filter as a loop, instead of having to throw a TypeError?

P-bibs:
This all looks good to me. Can we merge it?

PR #5, opened 2021-05-10
Add better tooling

P-bibs:
The goal of this PR is to add a few tools that make it easier for us to work on this repo:
* linting with ESlint to make sure we're following best practices and catching simple bugs
* formatting with Prettier so we don't have to worry about style convecntions
* CI with Github actions to make sure that PRs pass linting/formatting before they can be merged

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,19 @@
+# copied from gitignore
+
+/node_modules
+/.pnp
+.pnp.js
+
+/coverage
+
+/build
+
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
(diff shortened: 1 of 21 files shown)


P-bibs:
The formatter ran on a bunch of the code we'd already written so the diff is pretty messy, but most of the important changes in this PR are in `.eslintrc.js`, `package.json`, and `.github/workflows/node.js.yml`. What do you guys think @stardust66 @thomascastleman ?

P-bibs:
Also installing the Prettier and ESLint vscode extensions should hook into the config files without configuration to make formatting and linting work. Not sure about vim 😬

PR #6, opened 2021-05-10
add strings, logical not

thomascastleman:
This adds support for string literals to the language, as well as a logical not as a prefix unary operator that expects a strictly boolean operand. 

BTW this was branched off of the UI and CODAP integration branch, not main.


stardust66:
@P-bibs @thomascastleman are you good with merging this?

PR #10, opened 2021-05-10
Fix prettier

stardust66:
On my machine, `npm run format` wasn't targeting nested subdirectories, and adding single quotes around the file pattern globs in `package.json` following the suggestion from https://prettier.io/docs/en/cli.html fixed it.


stardust66:
I do have a vim plug-in that uses prettier, but I don't have one for eslint but that's okay. This PR purely affects the npm script though. Is this an issue on your end too or just mine?

P-bibs:
Yep, now that I look this is happening for me too (it's missing the __tests__ directory, for instance). Nice catch!

thomascastleman:
Good to merge?

PR #11, opened 2021-05-11
Refactor codapPhone and Transformation

stardust66:
### Changes
- Move types and enums to different file
- Resolve React hook missing dependency warning for the `useEffect` that listens to data updates. This makes sure that we are not re-registering a listener on every render, but only when one of our input parameters have changed.
- Move logic for fetching a list of data contexts into its own hook. This can go to a different file if that makes sense.


src/utils/codapPhone/types.ts
```diff
@@ -0,0 +1,95 @@
...
+  values?: any;
+}
+
+export interface CodapResponse {
+  success: boolean;
+}
+
+export interface CodapResponseValues extends CodapResponse {
+  values?: any;
+}
```
P-bibs, line 33: I think we may eventually want to fully annotate the structure of all codap objects (ex: what fields can a data context have, what fields can a collection have, etc.) which will let us make this type more concrete.

thomascastleman:
Looks good--does this also add support for automatically updating the transformed table when the input context is updated? Or what is happening with `addContextUpdateListener`?

PR #12, opened 2021-05-11
add row() built-in for accessing tricky attribute names

thomascastleman:
This adds some support for built-in functions, in particular, a `row()` function which takes a string name of an attribute. This allows us to access attributes that have spaces/other characters not valid in our language's identifier names.

For example: 
```
row("Long attribute name") > 100
```
as opposed to 
```
Long attribute name > 100
```
which will fail to parse.

### DIFF
--- a/src/language/parselets.ts
+++ b/src/language/parselets.ts
@@ -1,5 +1,5 @@
 import { Token } from "./lex";
-import { Ast, Operator, UnaryOperator } from "./ast";
+import { Ast, Operator, UnaryOperator, Builtin } from "./ast";
 import { getBindingPower, parseExpr } from "./parse";
 
 export interface PrefixParselet {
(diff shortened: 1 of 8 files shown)


src/language/parselets.ts
```diff
@@ -109,3 +109,51 @@ export class OperatorParselet implements InfixParselet {
...
+  parse(tokens: Token[], current_token: Token): Ast {
+    if (current_token.kind != "IDENTIFIER") {
+      throw new Error("Tried to use BuiltinParselet on non-identifier");
+    }
+    const name = current_token.content as Builtin;
+
+    const lparen = tokens.pop();
+    if (!lparen || lparen.kind !== "LPAREN") {
+      throw new Error(`Expected parenthesis after built-in "${name}"`);
+    }
```
P-bibs, line 123: I think this "expect and consume a certain token" functionality is common enough that we could break it off into a helper function.

P-bibs (commented):
Looks good!

One thing I should have mentioned is there's a sense in which you can parse function calls as an infix operation instead of a prefix operation. If you encounter a left paren in infix position, you know that what you've accumulated so far in `left_node` is an AST that evaluates to a function, and then you can parse your args to the right and stop when you hit a right paren. However, I don't think we'll notice any difference with this approach since we're not allowing user defined functions.

P-bibs:
Me as well! Good to merge?

PR #13, opened 2021-05-11
Use typescript version of no-unused-vars

stardust66:
See: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md

The plain `no-unused-vars` incorrectly warns against enum elements and argument names within interface definitions, etc. See [here](https://github.com/P-bibs/CODAP-flow/runs/2557748946#step:5:73). We are already using `@typescript-eslint/no-unused-vars`, which has the correct behavior.


P-bibs:
Good catch :+1:

PR #14, opened 2021-05-12
Better framework for naming of case tables/data contexts

P-bibs:
This PR includes some changes to the way our `codapPhone` api functions are setup to hopefully make managing names easier. 

The only names we need to worry about keeping unique are case tables and data contexts. With this in mind, I changed the way we generate a new table with data in `Transformation.tsx`. Rather than call a "create data context" and "create table" function manually, you call one function that does both, and pass a desired name you'd like to use (or null if you want one generated). This lets us ensure the name isn't already in use for a table/context before calling out to iframephone.

This also gives us a little more sanity for dev purposes, as now all collections/tables/contexts that are related will have a common number in their name, rather than gensyming a new name for each resource and having `table_1` contain `context_2` which has `collection_3`.


thomascastleman:
Thanks, this is great and will make naming much more sane. 

One thing I notice is that upon creating a table (either through another plugin or ours), the name shows up in our data context list as what appears to be the data context title (e.g. US Microdata). However, after producing a transformed version of that table, the original table then appears under its *case table* title. Any idea why this might be happening?

I feel like case table names are preferred for the dropdown, since they may correspond more intuitively to users' expectations of how things are named/grouped.

stardust66:
> One thing I notice is that upon creating a table (either through another plugin or ours), the name shows up in our data context list as what appears to be the data context title (e.g. US Microdata). However, after producing a transformed version of that table, the original table then appears under its _case table_ title. Any idea why this might be happening?

How are you getting this to happen? When I create a People table through the Microdata plugin, it always shows up in the dropdown as US Microdata, even after a transform.

thomascastleman:
@stardust66 Hmm interesting, I can get this by just creating the People table, then running it through a do-nothing transformation like filter `true`. The name changes from `US Microdata (US Microdata)` to `People (US Microdata)`. Same will happen with our tables if I transform them.

@P-bibs are you able to reproduce?

stardust66:
Oh nevermind I didn't catch that you meant on this branch.

P-bibs:
Hmm yes this also happens for me and I can't immediately understand why.

I agree that table names would make more sense for the user. However, are we sure that there's a one-to-one correspondence between tables and data contexts? I was under the impression that data contexts just sort of exist and you can display them through tables at will. This could lead to a scenario where there's two tables for one data context and we don't know which name to ues.

P-bibs:
Hmm still not sure I understand why it was using the collection name instead then but feel free to push

thomascastleman:
> However, are we sure that there's a one-to-one correspondence between tables and data contexts? I was under the impression that data contexts just sort of exist and you can display them through tables at will. This could lead to a scenario where there's two tables for one data context and we don't know which name to use.

Yeah, I'm not sure. I'm not against displaying both the case table and context name, as you're doing now, to disambiguate. It seems to let me rename two tables to the same name, but their context names are still unique and the user doesn't seem to have a way to change that.

This looks good to me!

stardust66 (approved):
Looks good to me too.

thomascastleman (approved):

PR #17, opened 2021-05-12
[WIP]: Use cases

stardust66:
Comparing with `complete-api-types-rebased` to make diff come out nicer. I'm only half way through but I seem to have run into some weird bug. I think the ideal case for us is actually to deal with data items when filtering, because it works like rows in a table would. Otherwise we would have to assemble the rows from the case hierarchy, which I haven't thought about too much yet. Anyway, right now I'm cloning the collection objects from the input data context and putting them in the new data context so the hierarchy would show up. However, when I insert data items after that, they seem to not show up in the table. The weird thing is that when you select the cases in the table, it shows presumably correct parent-child groupings, so there must be data in the new context. I have no idea why they aren't showing up.

image


thomascastleman:
Interesting, logging the data in the created table shows that the attributes seem to be there but values are all `undefined`.

Are we sure that we can add data to a data context that has multiple collections by supplying the data in item format? I just want to make sure this is possible because maybe there's another way we have to send along that kind of data if we want it to be slotted into the hierarchies of the data context.

> I think the ideal case for us is actually to deal with data items when filtering, because it works like rows in a table would. 

Yeah, this makes sense--it'd be nice though to always preserve the structure of the input. Maybe we could request data in item format specifically for filters, and piece back together the hierarchy before creating output? May be cumbersome. 

We might try to find a clever way to iterate over all "rows" in the collection even though we wouldn't have explicit rows when using the cases view. This code could be reused for a lot of transformations too.

stardust66:
Weird behavior, when I include the original id of the collection when copying it to a new data context, inserting works just fine but we get a Sproutcore warning that I assume would not be on the production version. IDs might have to be unique. I'm not sure why this makes insertion (via data items) work though.

image
image

P-bibs:
@thomascastleman I think that might work, but it might be hacky. From reading through the API I think it would look like this
1) do a wildcard `case` search to find all valid case ids for a context
2) do an `item` search for each of those case ids to make a table that allows us to translate between item and case ids
3) request all the items and filter them
4) use table from step 2 to resolve filtered items back to cases and add valid cases to new context

Hopefully there's a better way than this though.

src/utils/codapPhone/index.ts
```diff
@@ -156,28 +157,88 @@ export function getDataFromContext(
...
+        type: attr.type,
+        colormap: attr.colormap,
+        description: attr.description,
+        editable: attr.editable,
+        formula: attr.formula,
+        hidden: attr.hidden,
+        precision: attr.precision,
+        unit: attr.unit,
+      };
+    });
```
thomascastleman, line 210: @stardust66 @P-bibs if either of you have a better way of doing this feel free to change--I am trying to make sure only the fields outlined in the `BaseAttribute` type actually get copied into the new attributes.

The objects in `c.attrs` given to us by CODAP have more fields that we want to avoid copying.

PR #18, opened 2021-05-13
change `row()` to `attr()`, improve some error messages

thomascastleman:
I realized `row()` should probably be called `attr()` and several other places where we refer to "columns" should be in terms of "attribute" instead, to keep better consistency with the CODAP lingo.

I also updated a few error messages.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -133,7 +133,9 @@ function Transformation() {
           const result = evaluate(transformPgrm, dataEnv);
 
           if (result.kind !== "Bool") {
-            setErrMsg(`Expected boolean output, instead got ${result.kind}.`);
+            setErrMsg(
+              `Expected filter condition to evaluate to true/false, instead got a ${result.kind}`
+            );
             return;
           }
           // include in filter if expression evaluated to true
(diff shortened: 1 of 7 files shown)


stardust66 (approved):
Looks good. This makes sense.

thomascastleman:
@stardust66 @P-bibs Oops I just realized that their expression language handles the use of non-identifier names by wrapping it in backticks, i.e. 
```
`Long attribute name`
```
Should I update ours to use the same syntax, instead of having it be a built-in function kind of construct? It may be best to mirror their language for the most part.

P-bibs:
It does seem like consistency would be best. The only issue is now we have to add support for escaping backticks within those attribute names but maybe we can leave that for later

thomascastleman:
@P-bibs Hmm, if I use backticks within an attribute name it seems to give a syntax error (in their expression language) whenever I try to reference it, so maybe we don't need to handle backticks as part of attributes.

thomascastleman:
Okay, now we can do identifiers in backticks, like the CODAP expression language. It is effectively just a change in the lexer, where they are lexed as normal identifiers.

I left in support for built-ins since I figure we'll want that eventually, but removed `attr()` and just left an example built-in `isNegative()` that computes whether a number is negative, as a placeholder.

P-bibs:
Ahh sorry I should have been more clear. I think if you have an attribute name 
```
Bad`Name
```
then you can reference it in a formula as
```
`Bad\`Name`
```
At any rate, I think this is a good solution.

thomascastleman:
Ah, I see. Yeah, it might not be worth supporting right now if we may be using their language eventually anyway.

This otherwise good to merge? @P-bibs @stardust66

PR #20, opened 2021-05-14
abstract over transformations

thomascastleman:
This adds very preliminary support for multiple transformations, by extracting the functionality of filter out of the UI code. This is meant as a jumping off point and is WIP. 

This defines the notion of a "dataset" as:
```ts
export type DataSet = {
  context: DataContext;
  records: Record<string, unknown>[];
};
```
and a transformation as something which fulfills the following contract: 
```ts
export interface Transformation {
  (dataset: DataSet, extra?: unknown): DataSet;
}
```
That is, a transformation produces a whole new dataset as its output. This allows for transformations that may change not just records but the structure of the collections themselves (adding/deleting attributes, etc). The `extra` is intended to allow for extra information a transformation might need (e.g. in the case of filter, the predicate to filter by).

I've ported filter to this format (in `src/transformations/filter.ts`) and connected it to the UI to get an idea of how this might work.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -1,41 +1,9 @@
 /* eslint use-isnan: 0 */
 import React, { ReactElement } from "react";
-import { useState, useEffect, useCallback } from "react";
+import { useState } from "react";
 import "./Transformation.css";
 import Error from "./Error";
-import {
-  getAllDataContexts,
-  getDataFromContext,
-  setContextItems,
-  addNewContextListener,
-  removeNewContextListener,
-  addContextUpdateListener,
-  removeContextUpdateListener,
-  createTableWithData,
-  getDataContext,
-} from "./utils/codapPhone";
-import { Value } from "./language/ast";
-import { Env } from "./language/interpret";
-import { evaluate } from "./language";
-import { CodapIdentifyingInfo } from "./utils/codapPhone/types";
-import { filter } from "./transformations/filter";
-
-function useDataContexts() {
-  const [dataContexts, setDataContexts] = useState<CodapIdentifyingInfo[]>([]);
-
-  async function refreshTables() {
-    setDataContexts(await getAllDataContexts());
-  }
-
-  // Initial refresh to set up connection, then start listening
-  useEffect(() => {
-    refreshTables();
-    addNewContextListener(refreshTables);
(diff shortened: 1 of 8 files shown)


src/Transformation.tsx
```diff
@@ -151,11 +109,11 @@ function Transformation(): ReactElement {
...
-          setContextItems(lastContextName, newData);
+          setContextItems(lastContextName, filtered.records);
         } else {
           const [newContext, _newTable] = await createTableWithData(
             inputDataCtxt,
-            newData
+            filtered.records
           );
           setLastContextName(newContext.name);
         }
```
thomascastleman, line 119: Note that we'll probably need to change the way we generate transformation output, as opposed to just setting context items or creating a new table with the new items, because transformations may change the structure of collections.

On the CODAP phone end, we should probably add something like `createDataset` which creates a new data context of the given structure, and then also adds all the given items into it (as opposed to assuming only the items change and the rest can be cloned from the input data context by name). Does that seem reasonable? @stardust66 @P-bibs

thomascastleman:
Just realized the interface for transformations forbids transformations that might operate on multiple datasets (think diff, or.. merge?). Perhaps we want to expand it.

stardust66:
Hmm, I want to avoid using `unknown` and casting to deal with extra arguments. Since the transformations can get pretty different, can we actually write any code against the interface that works for all transformations?

I'm also thinking we might write different React components for each of the transformations because they might all have different UI. We can have a wrapper with a dropdown for transformation selection, and then use conditional rendering to render the correct transformation component below the dropdown. Each component can just import the transformation function it needs and there wouldn't have to be a generic `Transformation` interface.

src/Transformation.tsx
stardust66: Yeah, that makes sense. Maybe we'd want to set up some kind of diff so that we know which attributes are new and can add them. We want to clone old attributes sometimes, like if an attribute is computed, we want to make sure the formula is copied. This introduces complications like if the user changes a computed column with map, we need to make sure we don't copy the old formula.

thomascastleman:
I think maybe this is ready to merge as we're going to start adding more transformations from this base? What do you all think? @stardust66 @P-bibs

stardust66:
I agree

PR #23, opened 2021-05-18
switch from using `selected` on options to `defaultValue`

thomascastleman:
This gets rid of the warning when running tests, by using the `defaultValue` property to indicate which `<option>` the selects should have selected by default ("Select a Transformation" and "Select a Data Context"). 

I think this is fine, let me know if there's a better way to remove this warning.


P-bibs:
I'd never noticed this issue before but yep looks like defaultValue is best. LGTM :+1:

stardust66:
Looks good to me too!

PR #24, opened 2021-05-18
Add Diff component

P-bibs:
This PR adds a Diff transformation. Some notes:
* There are two kinds of diffs: categorical and numerical.
  * Categorical works like a diff on the lines of a file, where the algorithm tries to figure out which items have been added or removed between the two columns. I honestly don't think this one is very helpful right now but it seemed worth including as a proof of concept.
  * Numerical diff assumes both columns are numerical and computes their difference.
* This is definitely a component that would benefit from improved styling but I think we're trying to leave that until later. 

And some items I would appreciate thoughts/feedback on:
* Diff is definitely not the word we want to use in general. For someone without past experience, "compare" or "difference" make much more sense. Any thoughts on good names/wording?
* I was having trouble coming up with good test data. Feel free to play around with it and try to find edge cases.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "@types/node": "^12.20.12",
         "@types/react": "^17.0.5",
         "@types/react-dom": "^17.0.3",
+        "diff": "^5.0.0",
         "iframe-phone": "1.1.3",
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
@@ -22,6 +23,7 @@
         "web-vitals": "^1.1.2"
       },
       "devDependencies": {
+        "@types/diff": "^5.0.0",
         "@typescript-eslint/eslint-plugin": "^4.22.1",
         "@typescript-eslint/parser": "^4.22.1",
         "eslint": "^7.26.0",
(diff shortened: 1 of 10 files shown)


P-bibs:
What a fool I was to think that I could rely on truthiness and falsiness without being stabbed in the back. Good catch!

Yep that makes sense about the attributes. I had already laid some groundwork for deduplicating names earlier so I'll try to hook into that.

thomascastleman:
- This might be an interface question, in which a case we should probably punt until later when we have some feedback, but I wonder if the "no change" case for Compare Status should be colored in some way--right now if it's all no change, those cells just look blank, so people might get confused.

- One thing is that if you create an attribute that is numeric, and another that is a formula, then diff them as categorical (I know, this is bad to begin with), it seems to create extra rows:

formula-diff

Formulas seem to be the cause because this doesn't happen when doing a categorical diff on regular numbers. Ultimately they probably shouldn't be doing this but it might be kind of surprising if they do.

P-bibs:
* This makes sense. I originally had "no change" as grey but moved away from that. I can reinstate it
* It's actually very unclear to me what behavior is correct semantically here, and I think it's actually a good argument for copying concrete values instead of formulas. Just like we realized there's no guarantee that the output graph won't have changed collections or attributes from the input, it's also true that two datapoints which were in the same case may no longer be in the same case, which would make copying-by-formula break.
Then again, this categorical diff view isn't really a table at all, and maybe it's incorrect to change our invariants just for this one transformation.

thomascastleman:
I am also wondering how usable the output of a diff is as *data* for other transformations, or if it's really intended as more of a visual indicator. It might be nice if you wanted to diff and then filter by positive/negative changes--but is there currently a way to do this?

P-bibs:
I was actually thinking about that a lot. I think it's an important question to sit down and talk through at some point because if we decide it isn't useful as an input to other transformations, then we don't have to make it a CODAP table at all and could just make it a component of our own design with a much cleaner interface. 
If you wanted to filter by positive/negative changes I suppose you could filter on the difference category or the color category if we included constants like `RED` and `GREEN` in our language. Any other ideas?

thomascastleman:
Yeah that's a good point, worth asking maybe some teachers to see what the intended use cases of diff are, because having a custom component would be nice.

I suppose for numeric diffs at least you have the `Difference` attribute, which you could filter on directly to get negative / positive cases. And for categorical I'm not sure how much sense it would make to filter anyway.

PR #25, opened 2021-05-18
add group by

thomascastleman:
This adds a "group by" transformation which allows the user to select one or more attributes to then group a dataset by. The attributes are lifted out of their current collections into a new collection which becomes the parent of all. 

We just introduce a new collection with these attributes and CODAP handles the grouping for us.

Right now the "update current context" doesn't work because this is a transformation that makes a structural change to the context, so we need to make sure we are updating structure as well as items when we perform this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "@types/node": "^12.20.12",
         "@types/react": "^17.0.5",
         "@types/react-dom": "^17.0.3",
+        "diff": "^5.0.0",
         "iframe-phone": "1.1.3",
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
@@ -22,6 +23,7 @@
         "web-vitals": "^1.1.2"
       },
       "devDependencies": {
+        "@types/diff": "^5.0.0",
         "@typescript-eslint/eslint-plugin": "^4.22.1",
         "@typescript-eslint/parser": "^4.22.1",
         "eslint": "^7.26.0",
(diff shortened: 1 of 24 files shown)


P-bibs:
Two small things:
* Can we add a line that says to separate attributes by newlines in the textbox? That wasn't immediately apparent to me.
* Can we prefill the "Name of New Parent Collection" box with some default like "parent"

thomascastleman:
Yes, I do hope the UI will ultimately get to be something that's not just a textbox but I can definitely make those changes for now.

stardust66:
> Yes, I do hope the UI will ultimately get to be something that's not just a textbox but I can definitely make those changes for now.

I think in the future we could have dropdowns users could pick from. We can do that thing where we always create an empty one at the bottom whenever the user uses a dropdown. But this looks good for now!

stardust66 (approved):

P-bibs:
From Shriram: Make default parent name "Grouped By {attributeName}" @thomascastleman

thomascastleman:
This change preserves the original attributes after grouping, but adds grouped attributes to a new collection (as before), named appropriately to indicate they represent groups. This allows you to use the group-by to look *at the original data*, rather than having it alter the structure. Here's a preview:

preserved-attributes

The only thing that might get weird is now we're actually changing the records of the table (adding effectively copies of the original attributes), so if you did something like a flatten on the result, you'd get a different table (whereas what we had before just moved things into different collections, so flatten(original) == flatten(groupby(original))). I don't really think this is a problem though, because the copied attributes really are new attributes, so what you get from flatten is reasonable.

What do you all think? @P-bibs @stardust66

PR #26, opened 2021-05-18
Add aggregating operations (Fold)

stardust66:


thomascastleman:
I wonder if the new column for most of these transformations should be inserted into the collection that the input column came from, rather than always at the bottom level.

My thinking for this is you might have an attribute that's specifically about the parent, and you might want to do one of these folds (eg running sum of the parent atttribute), expecting it to attach onto the parent collection and give you info only about the parent. At the moment, this will compute the running sum over every child record of the parent, meaning the parent attribute will appear many times, which might be not what the user intended.

PR #27, opened 2021-05-19
add count

thomascastleman:
Adds a count transformation (as described [here](https://github.com/brownplt/B2T2/blob/main/TableAPI.md#count--t1table--ccolname---t2table)) which allows counting the frequency of distinct values under a given attribute.

This seems to work fine as an initial implementation, but I have a few thoughts: 

- It might be nice to have two "modes", one in which the counting is performed across the entire set of records (current implementation), and another in which it is performed locally for the records of each parent (ie, if the given attribute is part of a child collection, count within the records of the parents of that collection). I'm not sure what the output would look like for the latter.

- CODAP sometimes seems to use `""` to represent missing values, and count should probably take this into account. Right now, this shows up as a blank value in the values column, but we might consider displaying this with a better name ("N/A", or "Missing Value" maybe).

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -4,6 +4,7 @@ import { useState } from "react";
 import "./Transformation.css";
 import Error from "./Error";
 import { Filter } from "./transformation-components/Filter";
+import { Count } from "./transformation-components/Count";
 
 /**
  * Transformation represents an instance of the plugin, which applies a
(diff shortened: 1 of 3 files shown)


stardust66 (approved):

P-bibs (approved):
I think this is actually my favorite transformation yet since it's one of the only ones that provides functionality not already available in CODAP.

PR #28, opened 2021-05-19
Update context with dataset

stardust66:
I ran into some weird behavior when I tried to delete all collections and insert new ones from the dataset object when we do an update. So this is the same behavior as before, just using dataset instead.


thomascastleman:
Ah, I see. So this still doesn't update for structural changes like new/changed attributes or collections right?

PR #30, opened 2021-05-19
Add sort transformation

stardust66:
Asks user for a key expression which is used to sort the table.


thomascastleman (approved):
Nice, LGTM

P-bibs (approved):
Slick!

PR #33, opened 2021-05-24
Context update experiment

stardust66:
When updating a data context, this tries to first create an empty collection, delete all old collections, then add new collections and new data. However, it keeps running into a `TypeError: cyclic object value` on the CODAP side, which apparently happens when you try to `JSON.stringify` an object that has a self reference (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value). Despite this error, the operations do eventually complete, but it takes a while.


stardust66:
If you apply the following patch on the CODAP repo, the cyclic reference error gets fixed and our update happens a little bit faster, but it's a little bit too slow to be ideal.

```patch
From 6a106fc1bfa707d3d8b888216a9dbf617dd10d73 Mon Sep 17 00:00:00 2001
From: Jason Chen <kbtpodifo@gmail.com>
Date: Mon, 24 May 2021 11:52:06 -0400
Subject: [PATCH] Fix result from deleting collection

---
 apps/dg/controllers/document_helper.js | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)

diff --git a/apps/dg/controllers/document_helper.js b/apps/dg/controllers/document_helper.js
index e7fc4bdb..2812fa9d 100644
--- a/apps/dg/controllers/document_helper.js
+++ b/apps/dg/controllers/document_helper.js
@@ -215,11 +215,14 @@ DG.DocumentHelper = SC.Object.extend(
           return {success: false, values: {error: 'Collection not found'}};
         }
 
-        return context.applyChange({
+        var changeResult = context.applyChange({
           operation: 'deleteCollection',
           collection: collection,
           requester: iRequesterID
         });
+        return {
+          success: changeResult && changeResult.success,
+        };
       },
 
       createCollection: function(iResources, iValues, iID) {
-- 
2.30.2
```

stardust66:
I'm thinking maybe we can have two tracks for updates. If only the data has changed, then we can just update the cases, which is faster. Otherwise (if something is renamed, etc.), we update all collections.

stardust66:
This implements a comparison of the existing collections and the updated collections. If they are different, then we delete all the collections and add the new ones. If not, we only update the data.

I haven't extensively tested the comparison, but it seems to work. One problem is that CODAP modifies the collections object we give it. For example, if we give an object with a name but no title, CODAP will fill in the title field with the name. When we compare the new collections objects to the current version we query from CODAP, they will not be equal even though the collections have not changed. A solution could be to record the collections from the previous update and use that for comparison instead of querying from CODAP.

stardust66:
Replaced by #41

PR #34, opened 2021-05-24
Standardize component UI and functionality

P-bibs:
## Changelog
* Create a set of UI components for common UI elements (dropdown, text input, text area, transformation buttons) and switch over existing `transformation-components` to use the new UI components.
* Factor out repeated behavior from `transformation-components`. Affected behavior is as follows:
  * Listening for updates in source context and performing corresponding transformations: `useContextUpdateListenerWithFlowEffect`. This function takes a two context names and calls a callback when the first one updates. See the docstring for more info.
  * Deciding whether to update or create a new table and handling errors: `applyNewDataSet`. This function takes a dataset and some other callbacks/state and either makes a new table or updates an existing one, handling errors accordingly.
  * Turning a context name into a dataset: `getDataSet`. This function is just a small wrapper that uses CodapPhone requests to turn a context name string into a dataset.

## Notes:
The main point of this PR is to take functionality that's copy+pasted in different components and factor it out to a single function. Hopefully this will make it easier for us to change the corresponding functionality in the future, since there's only one place to tweak and all components get updated.

Because of this, there's some pretty verbose function argument lists/names. I erred on the side of taking more arguments rather than fewer, so if we need to add more functionality that requires the extra state we have it.

Finally, it's worth noting the UI components don't actually look any different, they're just in one place now. I suppose we should also do a styling PR at some point.

### DIFF
--- a/src/transformation-components/Compare.tsx
+++ b/src/transformation-components/Compare.tsx
@@ -1,16 +1,7 @@
-import React, { useState, useCallback, ReactElement, useEffect } from "react";
-import {
-  getDataFromContext,
-  setContextItems,
-  createTableWithDataSet,
-  getDataContext,
-  addContextUpdateListener,
-  removeContextUpdateListener,
-  getDataSet,
-} from "../utils/codapPhone";
+import React, { useState, useCallback, ReactElement } from "react";
+import { getDataSet } from "../utils/codapPhone";
 import {
   useAttributes,
-  useContextUpdateListener,
   useContextUpdateListenerWithFlowEffect,
   useDataContexts,
   useInput,
--- a/src/transformation-components/Compare.tsx
+++ b/src/transformation-components/Compare.tsx
@@ -1,13 +1,16 @@
 import React, { useState, useCallback, ReactElement } from "react";
 import { getDataSet } from "../utils/codapPhone";
 import {
-  useAttributes,
   useContextUpdateListenerWithFlowEffect,
-  useDataContexts,
   useInput,
 } from "../utils/hooks";
 import { compare } from "../transformations/compare";
-import { CodapFlowSelect, TransformationSubmitButtons } from "../ui-components";
+import {
+  CodapFlowSelect,
+  AttributeSelector,
+  ContextSelector,
+  TransformationSubmitButtons,
+} from "../ui-components";
 import { applyNewDataSet } from "./util";
 
 interface CompareProps {
@@ -31,10 +34,6 @@ export function Compare({ setErrMsg }: CompareProps): ReactElement {
     string | null,
     HTMLSelectElement
   >(null, () => setErrMsg(null));
-
-  const dataContexts = useDataContexts();
-  const attributes1 = useAttributes(inputDataContext1);
-  const attributes2 = useAttributes(inputDataContext2);
 
   const [lastContextName, setLastContextName] = useState<null | string>(null);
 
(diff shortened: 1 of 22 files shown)


src/ui-components/index.ts
```diff
@@ -0,0 +1,9 @@
+import a from "./CodapFlowTextInput";
+import b from "./TransformationSubmitButtons";
+import c from "./CodapFlowSelect";
+import d from "./CodapFlowTextArea";
+
+export const CodapFlowTextInput = a;
+export const TransformationSubmitButtons = b;
+export const CodapFlowSelect = c;
+export const CodapFlowTextArea = d;
```
P-bibs, line 9: Can someone show me a better way to do this? It seems like we should be able to do something like 
```js
import CodapFlowTextInput from "./CodapFlowTextInput";
import TransformationSubmitButtons from "./TransformationSubmitButtons";
import CodapFlowSelect from "./CodapFlowSelect";
import CodapFlowTextArea from "./CodapFlowTextArea";

module.exports = {
  CodapFlowTextInput,
  CodapFlowSelect,
  TransformationSubmitButtons,
  CodapFlowTextArea,
};
```
but that wasn't working.

src/transformation-components/SelectAttributes.tsx
```diff
@@ -62,26 +67,25 @@ export function SelectAttributes({
...
-      <textarea onChange={attributesChange}></textarea>
+      <CodapFlowTextArea onChange={attributesChange} value={attributes} />
 
       <br />
-      <button onClick={() => transform()}>Select attributes!</button>
+      <TransformationSubmitButtons
+        onCreate={() => transform()}
+        onUpdate={() => transform()}
+        updateDisabled={true}
+      />
```
P-bibs, line 88: Most of the transformation buttons aren't hooked up yet since most components don't have table update functionality. Eventually they will look like
```js
<TransformationSubmitButtons
onCreate={() => transform(false)}
onUpdate={() => transform(true)}
updateDisabled={!lastContextName} 
/>
```

P-bibs (commented):
Some initial notes

src/ui-components/index.ts
thomascastleman: Would something like we did [here](https://github.com/brownplt/duocode/blob/main/src/modules/index.js) work? Not sure if it's better but seems like it worked there.

P-bibs: Exactly what I was looking for, thanks!

P-bibs: Argh, actually I'm not sure this is quite right since then we can't quite use the `import { name } from "path"` syntax without an extra set of braces.

P-bibs:
@thomascastleman @stardust66 marking this as ready for review. I've tried to do some testing of every component but there could definitely be blind spots. I know there are still unfixed edge case bugs but I think as long as this PR doesn't introduce any new bugs we're good. Let me know if any of the changes need explanation or there's a better opportunity for abstraction somewhere.

src/ui-components/CodapFlowSelect.tsx
```diff
@@ -0,0 +1,35 @@
...
+  onChange,
+  value,
+  defaultValue,
+  options,
+}: CodapFlowSelectProps<T>): ReactElement {
+  return (
+    <select
+      onChange={onChange}
+      defaultValue={defaultValue}
+      value={value || defaultValue}
```
thomascastleman, line 23: What's the intention here with `value || defaultValue`? This seems to cause an error in the console, telling us to define either `defaultValue` or `value` but not both. I think maybe it's fine to just have `defaultValue` here?

src/transformation-components/Compare.tsx
```diff
@@ -97,112 +85,90 @@ export function Compare({ setErrMsg }: CompareProps): ReactElement {
...
-        {attributes1.map((attribute) => (
-          <option key={attribute.name} value={attribute.name}>
-            {attribute.title} ({attribute.name})
-          </option>
-        ))}
-      </select>
+        options={attributes1.map((attribute) => ({
+          value: attribute.name,
+          title: `${attribute.title} (${attribute.name})`,
+        }))}
```
thomascastleman, line 135: This title gives an odd doubling of the attribute name because the "title (name)" formatting is also happening in the CodapFlowSelect component. For example when I open this dropdown I see "State (State) (State)". We can probably just omit the formatting here and let CodapFlowSelect be the one to handle it.

See also lines 143-146 below this for the second attribute dropdown.

thomascastleman (commented):
Looks good to me other than the few changes mentioned 👍🏻

src/transformation-components/Compare.tsx
stardust66: I fixed this in #36 if you want to look at that

thomascastleman:
Also, is it worth refactoring the "choose a transformation" select as a `CodapFlowSelect`?
```tsx
<p>Transformation Type</p>
<select id="transformType" onChange={typeChange} defaultValue="default">
  <option disabled value="default">
    Select a Transformation
  </option>
  {Object.keys(transformComponents).map((type, i) => (
    <option key={i}>{type}</option>
  ))}
</select>
```
Seems like it would give us uniform styling, and also allow us to take advantage of the name/title separation (currently the transformation names we're presenting to the user look an awful lot like React component names lol)

PR #35, opened 2021-05-24
fix copying of attributes in flatten

thomascastleman:
Since every data context we get from CODAP passes through `normalizeDataContext`, its attributes are copied correctly and only include the fields we've defined in our `CodapAttribute` type.

Therefore, any further copying of attributes can simply do a normal copy, rather than having to use the `copyAttrs` functionality in `utils/codapPhone/index.ts`. I've updated flatten to no longer depend on this and removed the export of `copyAttrs`.


stardust66 (approved):
Looks good!

P-bibs:
LGTM

PR #36, opened 2021-05-24
Factor out context and attribute selectors

stardust66:
This just factors out context and attribute selectors into their own components.


P-bibs (approved):
This looks good to me. I believe it fixes most of the issues we had for #34

PR #39, opened 2021-05-25
add pivot longer/wider

thomascastleman:
Adds the pivot transformations detailed [here](https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html).

### DIFF
--- a/src/transformations/pivot.ts
+++ b/src/transformations/pivot.ts
@@ -1,0 +1,214 @@
+import { DataSet } from "./types";
+import { eraseFormulas } from "./util";
+
+/**
+ * Turns selected attribute names into values of a new attribute, reorganizing
+ * the values under the original attributes into a new column, thus making the
+ * dataset "longer" (more cases), but less wide (fewer attributes).
+ *
+ * @param dataset the dataset to pivot
+ * @param toPivot list of attribute names that should become values
+ * @param namesTo name of attribute under which toPivot names will go
+ * @param valuesTo name of attribute under which *values* of the toPivot
+ *  attributes will go
+ * @returns pivoted dataset
+ */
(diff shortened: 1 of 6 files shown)


thomascastleman:
Some examples:

longer

wider

stardust66:
Quick comment: It might be nice to have the same original rows be grouped together by default. I notice that when I pivot the grades example, rows with the same assessment names are grouped together, rather than rows that were produced from the same original row.

image

src/transformations/pivot.ts
```diff
@@ -0,0 +1,214 @@
...
+            `invalid attribute to retrieve names from: ${namesFrom}`
+          );
+        }
+        if (typeof rec[namesFrom] === "object") {
+          throw new Error(
+            `cannot use object values (${namesFrom}) as attribute names`
+          );
+        }
+
+        return rec[namesFrom] as string;
```
stardust66, line 110: Can we instead check

```typescript
if (!(typeof rec[namesFrom] === "string")) {
  throw new Error("Expected string, instead got ...");
}
```
Not a big deal, but want to avoid casting.

thomascastleman: I think we do want to allow things like numbers becoming attribute names though, right? E.g. if the numbers are really categorical in a sense. So maybe we should use `String(rec[namesFrom])` instead? @stardust66

stardust66: Oh that's right. Yeah, using `String(rec[namesFrom])` makes sense.

thomascastleman:
@stardust66 Agreed, and fixed!

stardust66 (approved):
Yay, looks good!

PR #41, opened 2021-05-25
Context update

stardust66:
This is the finalized context update PR, with the recent merged changes folded in. Now there are reactive updates when things other than data change, such as when the user renames an attribute. These collection updates are done by first adding an empty collection, then deleting all old collections, then adding all the new collections. Since this is more expensive than the usual update (takes longer, more visual flicker), we will check if the collections have changed before performing this kind of update. If only the data has changed, we just delete and insert the data items as before.

A caveat is that when we delete all collections, the table gets shrunk, which feels finicky and is not great user experience. Not sure how to fix this yet.

### DIFF
--- a/src/transformation-components/util.ts
+++ b/src/transformation-components/util.ts
@@ -1,6 +1,9 @@
 import { DataSet } from "../transformations/types";
-import { createTableWithDataSet, setContextItems } from "../utils/codapPhone";
 import { DataContext } from "../utils/codapPhone/types";
+import {
+  createTableWithDataSet,
+  updateContextWithDataSet,
+} from "../utils/codapPhone";
 
 /**
  * This function takes a dataset as well as a `doUpdate` flag and either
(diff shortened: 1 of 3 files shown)


thomascastleman:
Nice, this looks good. A few things: 

Does this implement the "if change is data-only, just update data items"? It seems that if I change data, the table still flickers when performing the update (as though it did the full delete).

One update that appears to not be caught by this currently is moving an attribute between collections. I'm not sure if we're able to listen for this, but it is a pretty big structural change that we might want to update on if possible.

stardust66:
> It seems that if I change data, the table still flickers when performing the update (as though it did the full delete).

Which transformation is it doing that for? I may have overlooked something. Comparing collections is a bit tricky because CODAP supplies default values for collection and attribute objects so when we query them we might get something slightly different to what we put in.

> One update that appears to not be caught by this currently is moving an attribute between collections. I'm not sure if we're able to listen for this, but it is a pretty big structural change that we might want to update on if possible.

I'll look into this.

thomascastleman:
Hmm, I'm not sure but it seems to be happening with build column. A change in one of the case cells alone seems to trigger a full update of the collections and not just data.

stardust66:
@thomascastleman Should be fixed now!

thomascastleman (approved):
Looks good!

PR #42, opened 2021-05-26
count with multiple attributes

thomascastleman:
This is a new version of count that allows the user to select multiple attributes to count by. It counts the frequency of every tuple of those attributes that is present in the input dataset. For instance, if you want to count the frequency of Sex/Age pairs, you could get the following output:

multi-count


stardust66 (approved):

stardust66:
Nice. Looks good!

P-bibs (approved):
LGTM

PR #43, opened 2021-05-26
name tables relative to transformations

thomascastleman:
This updates the naming of tables/contexts to use names relevant to the performed transformation, such as "Filter of ____" or "Count of ___".

Note that with this approach names can get pretty unwieldy pretty quickly, if you are heavily composing transformations.

table-names

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -56,7 +56,11 @@ function Transformation(): ReactElement {
       <Fold setErrMsg={setErrMsg} label="Running Max" foldFunc={runningMax} />
     ),
     "Running Difference": (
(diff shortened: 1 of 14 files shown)


stardust66 (approved):
This is great!

P-bibs:
This is really good and I think really increases the readability of the table names in the screenshot. However, I agree that as they get longer in may get unwieldy. Do you think it would help or hinder readability if we added some sort of punctuation for grouping? So instead of having 
`Compare of Flatten of Running Mean of Table1 and Difference From of Table2`
we could have
`Compare of [Flatten of [Running Mean of Table1] and [Difference From of Table2]`

P-bibs (approved):

PR #45, opened 2021-05-26
Add preliminary grouping of transformations

stardust66:
Screenshot from 2021-05-26 19-27-47

How should we group these transformations?

Note: this is a temporary interface. We can move to a nicer one in the future.


thomascastleman:
Woah nice, this already looks much better. Maybe we can discuss in our meeting tomorrow what would make the most sense as groupings for the target audience.

Also, it doesn't seem to render the corresponding transformation UI when I make a selection now--any idea why `typeChange` isn't getting triggered?

PR #46, opened 2021-05-27
Add "Categorical Compare" transformation

P-bibs:
This adds a new transformation under the compare component that implements the functionality we've talked about with comparing packets.
image

There are still many many edge cases to work out involving what happens when the two data contexts don't have the same set of attributes, or when two cases are identical besides the decision. Also, I think a better name is in order, but nothing comes to mind.

EDIT: name changed from "decision" to "categorical." The existing categorical diff was renamed to structural.

### DIFF
--- a/src/transformation-components/Compare.tsx
+++ b/src/transformation-components/Compare.tsx
@@ -134,7 +134,7 @@ export function Compare({ setErrMsg }: CompareProps): ReactElement {
         options={[
           { value: "categorical", title: "Categorical" },
           { value: "numeric", title: "Numeric" },
-          { value: "decision", title: "Decision" },
+          { value: "structural", title: "Structural" },
         ]}
         value={compareType}
         defaultValue="Select a type"
--- a/src/transformation-components/Compare.tsx
+++ b/src/transformation-components/Compare.tsx
@@ -4,7 +4,7 @@ import {
   useContextUpdateListenerWithFlowEffect,
   useInput,
 } from "../utils/hooks";
-import { compare } from "../transformations/compare";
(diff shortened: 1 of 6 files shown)


thomascastleman:
I think I am confused what this kind of diff is doing but: If I create two Microdata tables, then set all of the Sex cases in table A to "Male", and all the Sex cases in table B to "Female", then categorical compare, it still shows me Male/Male and Female/Female as the decisions. 

Shouldn't it show all tuples of Sex that it found in the tables (so just one, Male/Female) as the decisions collection?

stardust66:
Actually, I think the example is broken right now
image

P-bibs:
@thomascastleman @stardust66 should be fixed now. There was an error related to comparing against two columns with the same name.

thomascastleman:
Hmm now I get this result with blank cases when comparing Sex from two copies of the same table: 

compare

P-bibs:
Should be fixed now. The transformation in your screenshot probably still won't work since it involves comparing boundary objects for equality. Did we ever find a good solution for that?

There's still a remaining bug that I'm not going to worry about fixing with this PR since the proper behavior isn't immediately apparent. It happens when you compare between the same context and there's two cases that are equal for all attributes except the comparison attribute. An example is below:
image

thomascastleman (approved):
I have been using `JSON.stringify` for object comparisons in things like Count.

Ah I see, that's interesting. But yes, otherwise I think this looks good

P-bibs:
Good call, that seems to fix things. Look good to merge?

thomascastleman:
Yep, looks good to me.

PR #47, opened 2021-05-27
convert pivots/count to use multi-attribute select and new naming schema

thomascastleman:
Pivot / multi-attribute count were developed without the new names ("Pivot of ____") and the multi-attribute select component. This converts them to use those features in uniformity with the other transformations.


P-bibs (approved):
LGTM

stardust66 (approved):
Nice. Looks good.

PR #48, opened 2021-05-31
add join transformation

thomascastleman:
This adds a join transformation that behaves similarly to the drag-and-drop way of joining in CODAP.

CODAP allows you to perform a join by dragging one attribute from a context onto an attribute from another. In effect, it joins _collections_ together by copying attributes from the dragged attribute's collection (what I've called the joining collection) into the dragged-onto attribute's collection (what I've called the base collection). It fills in case data by searching the joining collection for the first case for which the dragged-from and dragged-onto attributes' values match. If no such match is found, the case values for the new attributes are left missing.

CODAP's join effectively mutates the base collection, by literally adding attributes that then use a formula + lookup function to reference the data from another context/collection. Our join copies their functionality but produces a new table representing the join. See the following example:

join-example

Worth noting is that the base attribute and joining attribute ("State" and "State" here) are both present in the joined table. I figured this is useful since you can decide to delete one of them later if necessary, but they may be named different things and indeed you may want to use formulas/transformations that reference one or the other.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -23,6 +23,7 @@ import {
 } from "./transformations/fold";
 import { PivotLonger } from "./transformation-components/PivotLonger";
 import { PivotWider } from "./transformation-components/PivotWider";
+import { Join } from "./transformation-components/Join";
 
 /**
  * Transformation represents an instance of the plugin, which applies a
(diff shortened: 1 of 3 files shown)


P-bibs (approved):
All looks good to me on the technical side of things.

I think whether or not to leave both columns in the result is an interesting question. It's not what I would first expect, since it doesn't match the CODAP behavior, but it also definitely makes more sense within the contexts of immutable transforms. I'm going to add this to the doc as a question to ask Emmanuel.

thomascastleman:
Data for good join example with birds:
```
State,State Bird
Nevada,Mountain bluebird
Oregon,Western meadowlark
Missouri,Bobwhite quail
Washington,Willow goldfinch
Virginia,Northern cardinal
Rhode Island,Rhode Island Red
```

PR #49, opened 2021-05-31
use attribute selectors instead of text inputs

thomascastleman:
This just converts a few transformations that were still using text boxes for user input of existing attribute names to now use the attribute selector component.

Also, I changed the fold transformations' naming so that their names appear capitalized in titles (e.g. "Running Sum of People").


stardust66 (approved):
LGTM

P-bibs (approved):
:+1:

PR #51, opened 2021-05-31
Trigger flow update on collection hierarchy update

### LINKED ISSUE #50: dragging attributes to new collection does not trigger update
If you drag an attribute of an input table so that it creates a new collection with just itself in it, a table that was produced by transforming that input will not be updated accordingly.

stardust66:
Fixes #50


thomascastleman (approved):
Nice!

PR #55, opened 2021-06-01
add a collection selector

thomascastleman:
Adds a dropdown selector for collections much like the existing `ContextSelector` and `AttributeSelector`. This is used exclusively in build column at the moment but could be useful in the future. Also fixed up a few names that seemed to be left over from prior copying of code.


P-bibs (approved):
:+1:

stardust66 (approved):

PR #57, opened 2021-06-01
improve appearance of names in dropdowns

thomascastleman:
This alters `CodapFlowSelect` so that it only displays the option's underlying value if there are multiple options that share the exact same title. I.e., in a situation in which the value is necessary to disambiguate between like titles. This greatly reduces the length of the dropdowns and removes clutter in most situations (when titles are indeed unique).

I also consolidated the "generate unique name" functionality that was required in both codapPhone utils and certain transformations under one function instead of having multiple that effectively do the same thing.

### DIFF
--- a/src/transformations/count.ts
+++ b/src/transformations/count.ts
@@ -1,6 +1,7 @@
 import { DataSet } from "./types";
 import { CodapAttribute, Collection } from "../utils/codapPhone/types";
-import { eraseFormulas, uniqueAttrName } from "./util";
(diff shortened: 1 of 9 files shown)


P-bibs:
Definitely makes the interface slicker, so possibly a good thing to merge soon.

I did notice that I had to remove line 27 of `CollectionSelector.tsx` to get it to run. Should I commit that change?

thomascastleman:
Yes please, that must've gotten overlooked when I merged to resolve conflicts

PR #60, opened 2021-06-01
Add ability to save transformations

P-bibs:
This PR adds the ability to save custom transformations based on a primitive transformation. All data about the transformation is saved except for the data context name.

## Known bugs:
* saved transformations with attribute selector doesn't display properly until a data context is selected


P-bibs:
Agree about editing and deletion. Maybe you get a `save` button on primitive transforms but it's replace with a `delete` and `edit` button on user-defined transforms?

Do you think we should have a chat with the 3 of us to discuss the single-window thing or bring it straight to sk?

thomascastleman:
I think it's probably good to go ahead and discuss in our meeting this morning

stardust66 (commented):
Otherwise looks good.

thomascastleman:
I see the note about displaying of attribute selectors in saved transformations not working yet, which is okay, but it seems to be more than a display problem:

Screenshot from 2021-06-04 16-29-17

It doesn't seem to remember the attributes I grouped by when I saved this transformation, and so errors. Is this the same on your end?

src/Transformation.tsx
```diff
@@ -98,7 +61,53 @@ function Transformation(): ReactElement {
...
+    // Create a new transformation and add it to the list
+    // TODO: can we do this without casting?
+    const content: SavedTransformationContent = {
+      base,
+      data,
+    } as SavedTransformationContent;
+
+    const savedTransformation = { name, content };
+    const encoded = encodeURIComponent(JSON.stringify(savedTransformation));
+    createDataInteractive(name, `http://localhost:3000?transform=${encoded}`);
```
thomascastleman, line 109: Thought: we may want the title of the spawned data interactive (currently "CODAP Flow") to related to the name of the saved transformation somehow. If a user has created a bunch of custom transformations, minimizing them is one way to keep the workspace from getting overwhelming, but then they all look the same. It would be nice if you could easily find a minimized transformation by its title.

src/ui-components/ExpressionEditor.tsx
```diff
@@ -122,20 +126,29 @@ export default function ExpressionEditor({
...
+        }}
+        onInputRead={(editor) => {
+          editor.showHint();
+        }}
+        onBeforeChange={(editor, data, value) => {
+          if (!disabled) {
+            onChange(value);
+          }
+        }}
+        onChange={(editor, data, value) => {
```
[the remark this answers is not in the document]
P-bibs: Fixed

src/Transformation.tsx
P-bibs: This should be fixed now

stardust66 (approved):

PR #61, opened 2021-06-01
Update attribute selectors when attributes change (deleted or renamed)

stardust66:
If an attribute gets selected and then gets deleted, the attribute selector will reset. A multi attribute selector will filter out any invalid selected attributes when they change.

### DIFF
--- a/src/transformation-components/Compare.tsx
+++ b/src/transformation-components/Compare.tsx
@@ -26,14 +26,10 @@ export function Compare({ setErrMsg }: CompareProps): ReactElement {
     string | null,
     HTMLSelectElement
   >(null, () => setErrMsg(null));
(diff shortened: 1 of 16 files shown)


thomascastleman:
This looks great, but I'm noticing one odd behavior: if you delete an attribute that is the last in its collection, it appears to remain in the dropdown (in the multi- and single- selects) and then will only go away once you make some change to another attribute.

Any idea why this is going on?

stardust66:
Good catch. I'll look at it later tonight.

stardust66:
Just rebased on top of main. The change from #51 fixes this.

thomascastleman (approved):
Nice, looks good

P-bibs (approved):
👍

PR #66, opened 2021-06-03
add copy transformation

thomascastleman:
This adds a transformation that produces an identical copy of the input table. This could be desirable because CODAP doesn't seem to have any native copying feature.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -25,6 +25,7 @@ import {
 import { PivotLonger } from "./transformation-components/PivotLonger";
 import { PivotWider } from "./transformation-components/PivotWider";
 import { Join } from "./transformation-components/Join";
+import { Copy } from "./transformation-components/Copy";
 
 /**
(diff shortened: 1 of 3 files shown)


P-bibs:
How would we feel about including a toggle that lets you control whether or not formulas will be erased in the resulting table? On the one hand it might be useful, on the other it might not fit the idea of composing transformations.

Also, I think whether or not we flow transformations to the copy is an interesting question. Seems like the difference between `copy` and `duplicate`. I'll add it to the list of Emmanuel questions.

Regardless, all looks good to me.

P-bibs (approved):

thomascastleman:
For formulas, I think we'd want to ask or do user testing to see if that's desired.

Yes, I think by default we should and only if they indicate they want a snapshot would we sever the connection (just to keep the idea of snapshots not married to any one transformation). Though we may want a snapshot transformation in and of itself. Definitely something to discuss

PR #67, opened 2021-06-03
Make CODAP Flow

stardust66:
update

Changes
- Remove update buttons
- Make created tables always update

We need to figure out what to do with errors. Currently, any errors during updates crash the plugin. We need to figure out how to present errors to the user.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -17,7 +17,7 @@ import { createDataInteractive } from "./utils/codapPhone";
  */
 export const SaveTransformationContext = React.createContext<
   (name: string, d: TransformationSaveData) => void
->((name, d) => {
+>(() => {
   // If someone tries to subscribe to this context without a valid
   // provider higher up, then throw an error.
   throw Error("No Save Transformation Provider Found");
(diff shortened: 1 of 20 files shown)


stardust66:
I added error handling, but the error messages are quite long.
image

stardust66:
There's also the question of what to do on multiple errors. We need a better way to indicate errors than the one place where some error message might appear.

thomascastleman:
Wooo! This is great. I think we should discuss what to do about errors at the next meeting, or maybe even ask Emmanuel at some point.

This looks great, and if the errors are verbose at least they give enough info to solve the problem. One thing I do notice though is if I run into an error, then fix the problem (add back in a deleted attribute, etc), the message persists, even though the tables update successfully. It would be nice to have the error get cleared somehow

stardust66:
> if I run into an error, then fix the problem (add back in a deleted attribute, etc), the message persists, even though the tables update successfully. It would be nice to have the error get cleared somehow

The latest commit should fix this. I'm a little worried about errors overriding each other, but we can discuss that later.

PR #68, opened 2021-06-07
attach description of transformation to output table 

thomascastleman:
This uses the [undocumented `metadata` property](https://codap.concord.org/forums/topic/accessing-dataset-description-through-plugin-api/) of data contexts to set the description of the output table, which can be viewed by clicking the (i) button on the top right of the table's side panel:

i-button

Each transformation now generates a description of what that particular invocation of the transformation did to the input, and attaches this description to the output data context. If a user forgets exactly how a table was computed (ie if the name isn't enough or is ambiguous), they can click the (i) button and see a description like this:

descrip

This is mainly to explore this idea--a lot of the descriptions could use heavy improvement/readability. We also might run into issues with outdated descriptions to due a change in the transformation itself (maybe this isn't a problem if we're not allowing "update this table" after a change to the transformation), and unfortunately as Jonathan says you can't _update_ the metadata property.


thomascastleman:
@stardust66 @P-bibs I've updated this to work with the data-driven changes. Please have a look at the descriptions and let me know if anything needs to be clearer or is in error.

P-bibs (approved):
I'm actually really impressed with how readable the descriptions are even after you're a few transformations deep. The only thing tripping me up is attribute lists being pluralized even if there's only a single item, which is probably ok to leave.

stardust66 (approved):

PR #69, opened 2021-06-07
change fold names + use attribute descriptions

thomascastleman:
This updates the naming of fold operations to incorporate the name of the table on which they were computed in their attribute:

running-mean

The format is now "\<fold type\> of \<attribute\> from \<data context\>". If it detects a long attribute name (one that contains spaces), it will parenthesize the inner attribute.

The attribute `description` property (which appears in a tooltip) is used to provide a full description of the fold that took place. I've also incorporated this `description` property into other transformations that create new attributes or significantly modify them.

### DIFF
--- a/src/transformations/fold.ts
+++ b/src/transformations/fold.ts
@@ -8,7 +8,8 @@ function makeNumFold<T>(
   return (
     dataset: DataSet,
     inputColumnName: string,
-    resultColumnName: string
+    resultColumnName: string,
+    resultColumnDescrip: string
   ): DataSet => {
     let acc = base;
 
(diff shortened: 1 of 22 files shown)


src/transformations/fold.ts
```diff
@@ -8,7 +8,8 @@ function makeNumFold<T>(
   return (
     dataset: DataSet,
     inputColumnName: string,
-    resultColumnName: string
+    resultColumnName: string,
+    resultColumnDescrip: string
```
P-bibs, line 12: See above
```suggestion
    resultColumnDescription: string
```

P-bibs (approved):
See comments. This is really helpful, now we just have to remember to let users know they can hover on attributes to get this info so they don't totally forget about it.

stardust66 (approved):
Agree with everything @P-bibs said. Otherwise, looks great!

thomascastleman:
Okay, I've updated data contexts to use the same parenthesized names. I agree this helps with ambiguity.

PR #70, opened 2021-06-07
cover pyret table operations

thomascastleman:
This adds two transformations, "combine cases" and "copy schema", which mirror the `stack` and `empty` methods from the [Pyret tables documentation](https://www.pyret.org/docs/latest/tables.html).

Combine cases builds a table with the combined cases of two tables who share the same attribute set. Copy schema creates a copy of a table, but without any records (just the schema is copied).

With these added, we pretty much cover what I would say are the Pyret table operations that apply to our context. There may be more possibilities in the future with transformations that yield a single value, but for now I think this is good.

### DIFF
--- a/src/transformation-components/CombineCases.tsx
+++ b/src/transformation-components/CombineCases.tsx
@@ -8,8 +8,7 @@ import { applyNewDataSet, ctxtTitle, addUpdateListener } from "./util";
 import { TransformationProps } from "./types";
 import TransformationSaveButton from "../ui-components/TransformationSaveButton";
 
-// eslint-disable-next-line
-export interface CombineCasesSaveData {}
+export type CombineCasesSaveData = Record<string, never>;
 
 interface CombineCasesProps extends TransformationProps {
   saveData?: CombineCasesSaveData;
--- a/src/transformation-components/CombineCases.tsx
+++ b/src/transformation-components/CombineCases.tsx
@@ -1,0 +1,90 @@
+import React, { useCallback, ReactElement } from "react";
+import { getContextAndDataSet } from "../utils/codapPhone";
+import { useInput } from "../utils/hooks";
+import { combineCases } from "../transformations/combineCases";
+import { DataSet } from "../transformations/types";
+import { ContextSelector, TransformationSubmitButtons } from "../ui-components";
+import { applyNewDataSet, ctxtTitle, addUpdateListener } from "./util";
+import { TransformationProps } from "./types";
+import TransformationSaveButton from "../ui-components/TransformationSaveButton";
+
+// eslint-disable-next-line
+export interface CombineCasesSaveData {}
+
+interface CombineCasesProps extends TransformationProps {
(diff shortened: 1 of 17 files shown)


P-bibs (approved):
Looks good :+1: 

Also I'm glad someone found a use for those set util functions I made lol. It may make more sense for us to switch to a library like lodash for things like that if they become performance critical.

src/transformation-components/CombineCases.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useInput } from "../utils/hooks";
+import { combineCases } from "../transformations/combineCases";
+import { DataSet } from "../transformations/types";
+import { ContextSelector, TransformationSubmitButtons } from "../ui-components";
+import { applyNewDataSet, ctxtTitle, addUpdateListener } from "./util";
+import { TransformationProps } from "./types";
+import TransformationSaveButton from "../ui-components/TransformationSaveButton";
+
+// eslint-disable-next-line
+export interface CombineCasesSaveData {}
```
stardust66, line 12: I think we could do 
```suggestion
export type CombineCasesSaveData = Record<string, never>;
```
which is the canonical way to represent an empty object. See: https://github.com/typescript-eslint/typescript-eslint/issues/2063#issuecomment-675156492. Then we can get rid of `eslint-disable-next-line`.

src/transformation-components/CopySchema.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useInput } from "../utils/hooks";
+import { copySchema } from "../transformations/copySchema";
+import { DataSet } from "../transformations/types";
+import { TransformationSubmitButtons, ContextSelector } from "../ui-components";
+import { applyNewDataSet, ctxtTitle, addUpdateListener } from "./util";
+import { TransformationProps } from "./types";
+import TransformationSaveButton from "../ui-components/TransformationSaveButton";
+
+// eslint-disable-next-line
+export interface CopySchemaSaveData {}
```
stardust66, line 12: See above
```suggestion
export type CopySchemaSaveData = Record<string, never>;
```

stardust66 (approved):
Looks good!

src/transformation-components/CombineCases.tsx
thomascastleman: Unfortunately this seems to cause a type error and I'm not sure why:

Screenshot from 2021-06-08 10-56-28

stardust66: Huh, why is this happening in Compare...

thomascastleman: To be clear it happens in a lot of other transformations too once I change a few to use the `Record<string,never>`.

stardust66: I see. The type of `inputAttribute1` is `string | null`, but in the save data it's only allowed to be `string`.

PR #71, opened 2021-06-07
UI style

thomascastleman:
This is a WIP PR of adding some spit + polish to the UI before the demo on Wednesday. I've just started with the font and some of the buttons to look like other built-in plugins.

Feel free to just commit to this and we can get it to wherever we think is good enough before our meeting.


thomascastleman:
Screenshot from 2021-06-07 15-20-19

For reference so far

PR #72, opened 2021-06-08
Dot product and average

stardust66:
Until the bug on the CODAP side gets fixed, we can't create or update text through the API, so this just uses single-celled tables instead. For dot product, we could also have the user choose which columns to take the dot product of. Right now it takes the dot product of all columns.

### DIFF
--- a/src/Transformation.tsx
+++ b/src/Transformation.tsx
@@ -58,6 +58,7 @@ function Transformation({
       "Join",
       "Eval",
       "Copy",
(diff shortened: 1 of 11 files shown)


stardust66:
Alright text is working now with the workaround.

thomascastleman (approved):
I agree about wanting to select columns for the dot product, but this is good as a first version.

Looks good to me!

PR #73, opened 2021-06-08
disable browser launch on `npm start`

thomascastleman:
This makes it so that when you run `npm start`, it doesn't launch `localhost:3000` in the browser, since we are really never using the plugin as a standalone app.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+BROWSER=none


P-bibs (approved):
This looks good to me. It would probably be even better to `npm start` open the browser to `http://localhost:4020/dg?di=http://localhost:3000` but I'm not sure if there's a way to do that without ejecting from `create-react-app`.

PR #75, opened 2021-06-08
fixing single attribute selectors

thomascastleman:
This is a WIP to fix the single attribute selectors when saved. They previously weren't appearing as grayed out because the `disabled` prop wasn't getting passed through (this was happening in a few other places too which I've changed). 

However now there's the behavior of attribute selectors not holding their saveData value when they are used in a saved transformation. I logged `saveData` and then `attributeName` from inside the body of the `TransformColumn` component during the creation of a saved transformation and saw this:

Screenshot from 2021-06-08 19-11-10

Which is suspicious, given it's happening twice and the second time `attributeName` is `null` though it is present in the `saveData`.

Let me know if you have ideas about a fix for this / feel free to commit anything

### DIFF
--- a/src/transformation-components/Average.tsx
+++ b/src/transformation-components/Average.tsx
@@ -77,6 +77,7 @@ export function Average({ setErrMsg, saveData }: AverageProps): ReactElement {
         context={inputDataCtxt}
         value={attribute}
         onChange={setAttribute}
+        disabled={saveData !== undefined}
       />
       <br />
       <TransformationSubmitButtons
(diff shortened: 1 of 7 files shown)


stardust66:
Basically, to keep the attribute selection updated (i.e., if the chosen attribute gets deleted, we reset the dropdown), we use a `useEffect` to check if the chosen attribute is still an attribute of the selected data context. But in a saved transformation, the attribute needs to stay the same even when there's no data context, so I added a change that ignores the result of `useAttribute` if the selector is disabled.

PR #76, opened 2021-06-09
recategorize transformations + remove Eval

thomascastleman:
Adds some new categories so transformations are organized a bit more, and removes the demo Eval transformation since we have other transformations that are successfully using that functionality.

Screenshot from 2021-06-09 09-43-07


P-bibs (approved):
LGTM

PR #79, opened 2021-06-10
Add save data for dot product and fix graphical issue on multi attribute selector

P-bibs:
This PR
* adds save data for the dot product transformation
* fixes a graphical issue where disabled buttons in a multi-attribute selector would still highlight on hover and change the cursor type

### DIFF
--- a/src/transformation-components/DotProduct.tsx
+++ b/src/transformation-components/DotProduct.tsx
@@ -12,8 +12,9 @@ import { TransformationProps } from "./types";
 import TransformationSaveButton from "../ui-components/TransformationSaveButton";
 import { MultiAttributeSelector } from "../ui-components";
 
-export type DotProductSaveData = Record<string, never>;
-
+export interface DotProductSaveData {
+  attributes: string[];
+}
 interface DotProductProps extends TransformationProps {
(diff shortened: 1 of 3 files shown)


src/transformation-components/DotProduct.tsx
```diff
@@ -28,7 +29,11 @@ export function DotProduct({
     HTMLSelectElement
   >(null, () => setErrMsg(null));
 
-  const [attributes, setAttributes] = useState<string[]>([]);
+  const [attributes, setAttributes] = useState<string[]>(
+    saveData !== undefined ? saveData.attributes : []
+  );
+
+  console.log(attributes);
```
stardust66, line 36: Do we want to leave this in?

stardust66 (approved):

thomascastleman:
While you're at fixing dot product, could you also fix the `generateSaveData` function of Average? It has save data but returns `{}`.

src/transformation-components/DotProduct.tsx
P-bibs: Fixed

thomascastleman (approved):
Looks good

PR #80, opened 2021-06-10
add partition

thomascastleman:
This adds a partition transformation, which generates separate datasets for all distinct values of a given attribute in an input dataset.

partition

Updates currently work by maintaining which contexts currently exist as a result of the partition, and then updating existing contexts, adding new ones for new distinct values, and deleting contexts that no longer appear in the partition output.

### DIFF
--- a/src/transformations/partition.ts
+++ b/src/transformations/partition.ts
@@ -1,0 +1,56 @@
+import { DataSet } from "./types";
+
+/**
+ * Contains a dataset as a result of a partition, and the distinct
+ * value that all records of the dataset contain for the attribute
+ * by which partitioning was performed.
+ */
+export interface PartitionDataset {
+  dataset: DataSet;
+  distinctValue: string;
+}
+
+/**
+ * Breaks a dataset into multiple datasets, each which contain all
+ * cases with a given distinct value of the indicated attribute.
+ */
+export function partition(
+  dataset: DataSet,
+  attribute: string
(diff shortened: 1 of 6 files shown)


src/transformations/partition.ts
```diff
@@ -0,0 +1,58 @@
...
+  attribute: string
+): PartitionDataset[] {
+  // map from distinct values of an attribute to all records sharing that value
+  const partitioned: Record<string, [unknown, Record<string, unknown>[]]> = {};
+
+  const records = dataset.records.slice();
+  for (const record of records) {
+    if (record[attribute] === undefined) {
+      throw new Error(`Invalid attribute: ${attribute}`);
+    }
```
P-bibs, line 29: I noticed that if I apply a partition transformation, then delete the attribute that was partitioned on, I get a JS error instead of an error in our little dialog box. Is this expected behavior or something that needs fixing?

P-bibs (approved):
Really cool stuff. I'm especially impressed with how many edge cases work properly like deleting/adding new tables and partitioning by boundaries.

src/transformations/partition.ts
thomascastleman: Good catch. I wasn't sure why this was happening but I realized that the try/catch around the _creation_ of the update listener doesn't actually catch things that come from when that listener is actually run. Fixed in the latest commit.

PR #83, opened 2021-06-10
Add generic fold

stardust66:
image

Right now, we are having the user supply the name of the accumulator. If we always bind it to something like `acc`, then we might run into name conflicts if a column is also named `acc`. This avoids that but does add some complexity.

This "generic" version is also not very powerful. E.g., you can't write something like running average in it, because the expression language doesn't support tuples, so you can't accumulate multiple values that you can then use to calculate the output.

### DIFF
--- a/src/transformation-components/GenericFold.tsx
+++ b/src/transformation-components/GenericFold.tsx
@@ -1,12 +1,11 @@
 import React, { useCallback, ReactElement, useState } from "react";
 import { getContextAndDataSet } from "../utils/codapPhone";
-import { useInput } from "../utils/hooks";
+import { useInput, useAttributes } from "../utils/hooks";
 import { TransformationProps } from "./types";
 import { DataSet } from "../transformations/types";
 import {
   TransformationSubmitButtons,
   ContextSelector,
-  AttributeSelector,
   CodapFlowTextInput,
   ExpressionEditor,
 } from "../ui-components";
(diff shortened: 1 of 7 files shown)


src/transformation-components/GenericFold.tsx
```diff
@@ -0,0 +1,185 @@
...
+      return;
+    }
+    if (expression === "") {
+      setErrMsg("Please enter an expression");
+      return;
+    }
+    if (base === "") {
+      setErrMsg("Please enter a base value");
+      return;
+    }
```
thomascastleman, line 84: Hmm, if I'm doing a fold over strings and want the base value to be the empty string, this disallows that.

This makes me wonder if what they input for the base value should actually be a full expression that we evaluate to get the base value. This seems like it might be useful and would get around the above issue. Thoughts?

src/transformations/fold.ts
```diff
@@ -47,6 +48,42 @@ function makeNumFold<T>(
   };
 }
 
+export async function genericFold(
+  dataset: DataSet,
+  base: string,
+  expression: string,
+  inputColumnName: string,
```
thomascastleman, line 55: Do we need an `inputColumnName` at all? This made sense for other folds, since they were restricted to folding over a column. Since we're allowing the user to supply an expression, they can access the whole row, so I don't think we need them to indicate a particular column.

src/transformation-components/GenericFold.tsx
stardust66: I think that's a good idea. Then empty strings will be more explicit.

thomascastleman (approved):
Looks good, I just added some stuff about unique names with the result column.

PR #85, opened 2021-06-14
Add features to facilitate Bootstrap

P-bibs:
This PR adds two features that make it easier to follow the design recipe with CODAP-flow: purpose statements and type contracts.

Purpose statements allow you to add a description to a saved transformation (in addition to the existing title).

Type contracts allow you to add an intended output type to transformation formulas. This type is then dynamically checked when the transformation is applied and an error is thrown if the output types don't match the type contract. Supported types are `any`, `string`, `number`, `boolean`. Supported transformations are `sort`, `buildColumn`, `transformColumn`, and `sort`.

The type-checking is very permissive, since the data we get from CODAP is so unpredictable. Additionally, we'll probably want to move to a more CODAP-centric type system (think types like "categorical", "numeric", and "boundary").

## Misc notes:
* Also adds ascending/descending chooser to `sort`
* [Rewritten version](https://docs.google.com/document/d/1SdiV12kEwflf0miBsQgWszh376x4f_SCHepzSNb5NbE/edit?usp=sharing) of [table methods](https://www.bootstrapworld.org/materials/fall2020/en-us/courses/data-science/lessons/ds-table-methods/index.shtml) using CODAP-flow


thomascastleman (approved):
Looks good to me. 

I do think we'll want to discuss with sk / Emmanuel eventually about what exactly the typing rules should be that would make the most sense, but for a start this is good and we can easily change those rules later.

stardust66 (approved):
Nice. This looks good.

PR #87, opened 2021-06-17
add support for removing listeners after context deletion

### LINKED ISSUE #81: updates persist even after the context being updated is deleted
We currently don't have anything that says to stop listening for updates when a transformed table is deleted permanently. To reproduce:

1. Create a dataset
2. Transform it (e.g. create a Copy of Dataset)
3. Delete the transformed output (by clicking the trash icon in the tables menu, **not** by just clicking the X)
4. Make some change to the input dataset.

We try to propagate updates to the transformed dataset, but it no longer exists.

thomascastleman:
This, together with the [CODAP-side patch](https://github.com/stardust66/codap/commit/73212082dd7c6eae8346b6d44ff16217f83c8c03) that adds notifications for deleted contexts, allows us to deactivate any listeners that rely on a context existing when it is destroyed.

I added a notion of "dependencies" to the map from context names to listeners. When listeners are added, you indicate an array of context (or other component) names that identify which contexts must still be valid for the listener to be able to run. When a context is deleted, we remove any listeners on it as well as listeners that depend on it.

This fixes #81.

### DIFF
--- a/src/transformation-components/Partition.tsx
+++ b/src/transformation-components/Partition.tsx
@@ -75,31 +75,38 @@ export function Partition({
     try {
       const transformed = await doTransform();
       let valueToContext: Record<string, string> = {};
+      const outputContexts: string[] = [];
 
       for (const [partitioned, name] of transformed) {
         const newContextName = await applyNewDataSet(partitioned.dataset, name);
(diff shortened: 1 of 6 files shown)


stardust66 (approved):
Looks good!

PR #89, opened 2021-06-18
Deep copy collections and attributes in transformations

stardust66:
Since `array.slice()` only shallow copies the array, we are actually mutating records and collections in some of our transformations, which interacts badly with caching. This PR replaces all slices with actual copies. In the case that we don't mutate the objects, the slices have been removed, since they are misleading.


thomascastleman (approved):
Looks good

P-bibs (approved):
:+1:

PR #91, opened 2021-06-21
better errors for boundary objects

thomascastleman:
This makes our CODAP value printer aware of boundary objects (by looking for the property `jsonBoundaryObject` within them) and fixes pivot wider to give a more specific error message when boundary objects are used as attribute names.

This does beg the question of how many ways objects can show up as CODAP values, and it does seem like it's limited to boundary objects and a few others. I did find that using just the identifier `US_county_boundaries` as a formula value, for instance, yields an object that has a map from county names as keys to boundary objects as values, so this may be another kind of object we want to be aware of.

### DIFF
--- a/src/transformations/pivot.ts
+++ b/src/transformations/pivot.ts
@@ -1,5 +1,5 @@
 import { DataSet } from "./types";
-import { eraseFormulas, codapValueToString } from "./util";
+import { eraseFormulas, codapValueToString, isBoundary } from "./util";
 
 /**
  * Turns selected attribute names into values of a new attribute, reorganizing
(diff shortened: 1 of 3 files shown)


thomascastleman:
Okay, I added support for detecting things like `US_county_boundaries`, calling them "boundary maps" though I don't know if this is the best name.

P-bibs (approved):
Looks good to me :+1:. I think there's one more section that could benefit from a switch over to `isBoundary`/`isBoundaryMap` at `src/utils/prettyPrint:11`

stardust66 (approved):
Looks good to me!

PR #92, opened 2021-06-21
add warnings to partition on large outputs

thomascastleman:
This adds warnings (via `confirm()`) when you try to partition an attribute that has many distinct values. It also warns when you apply an update to the partitioned attribute that would cause the number of output datasets to be large. See below:

warn-create

warn-update

Note that if Cancel is chosen in the update case, the update still goes through to the input table, but the output tables just do not see the change. This might lead to some weird/inconsistent state.


P-bibs (approved):
Looks good to me. Good note about the weird behavior in the update case, although my guess would be that that specific path of execution will be encountered relatively infrequently so we don't have to worry about small inconsistencies.

PR #93, opened 2021-06-22
Convert transformations to data-driven

P-bibs:
Provides a mechanism to make transformations from data rather than manually creating each component. Also converts existing components to this new system.

Transformations can now be added by contributing to `transformationList.ts`. Each transformation can use up to 2 of the following UI components:
* context selector
* attribute selector
* collection selector
* multi-attribute selector
* text input
* text area
* dropdown
* expression editor
* type contract

Each of these components must have at least  a `title` provided, but some require additional configuration (like the `option`s for a dropdown).

## Implementation Notes
The `DataDrivenTransformation` component contains the state for all possible UI elements (2 of each in the list above). The main difficulty is in going from this large set of state to the transformation functions that operate on datasets (like `buildColumn`, `filter`, etc.). To that end, each file in the `transformations` folder now has a shim function. It takes a `DDTransformationState` and pulls out the bits that are important to the given transformation (for instance, `filter` pulls out a context and an expression) to make sure they aren't null or invalid. It then calls the actual `DataSet`-producing functions, which have been renamed to have a prefix of `unchecked`.

The goal of this is to maintain a disconnect between error checking and functionality. the `filter` function is now only responsible for error checking the inputs, while the `uncheckedFilter` function implements the actual filter functionality.

## Followup
There are a few things we can restructure now to make the code cleaner. For instance, I want to rename `PolymorphicComponent.tsx` since it's a pretty ambiguous name. Also, I think we can start making `transformationList.ts` a single source of truth for transforms. For example, we could move the grouping for the dropdown to `transformationList.ts` instead of `Transformation.tsx`. However, this PR is big enough as is, so I'll get to work on these other changes as soon as we merge this one.


stardust66:
> Also, I think we can start making `transformationList.ts` a single source of truth for transforms. For example, we could move the grouping for the dropdown to `transformationList.ts` instead of `Transformation.tsx`.

Agreed. Then we'll only have to make changes in a single place when adding new transforms.

thomascastleman:
This is great and agreed about all of these things.

I am noticing that when you switch between transformations that share parts of their UI, those components aren't re-rendered/reset (which is of course what we expect from React). This is a little weird though, because if I fill out attributes or an input data context for one transformation, then switch to another, those inputs might persist (if they're shared by both transformations), even though they might not make any sense for the newly-selected transformation.

Is there a way we could prevent this? Somehow force the components to reset when transformations are selected?

stardust66:
> Is there a way we could prevent this? Somehow force the components to reset when transformations are selected?

I think we can move the state reducer into `Transformation.tsx` and just pass `state` and `setState` to `DataDrivenTransformation`. Then we can reset the state when a new transformation gets selected.

P-bibs:
@thomascastleman @stardust66 Pushed a small change. Does this fix the behavior you were observing?

stardust66:
Yup. That does it.

stardust66:
I noticed that the whitespace between the expression editor and the type contract has disappeared. In the old version there used to be a `<br />` there. I think a fix would be to replace the `<></>` wrapping each input section with divs that set the top margin. Then we can stop setting `margin-top` in `h3`s so that sections without titles can have the same spacing.
image

This should probably be in another PR.

thomascastleman:
That fixes the behavior for the transformation inputs, but I am observing the same thing happening for save name/purpose statement. It would be great if we could clear those as well.

Perhaps more importantly, it looks like save data is broken now--I'm not sure if it was working before the fix as I didn't check before. The saved windows have disabled but empty inputs.

P-bibs:
Save data should be working now.

I'd actually argue that it's ok for the title/purpose statement not to be erased when the user switches transformations. However, I have changed it so that they're erased when the user hits save. Is that acceptable?

stardust66:
I think that works.

thomascastleman:
Yes, that can work.

Okay one final thing: Right now if you select an input table and save a transformation, the saved one pops up with that input table still selected. It feels like it should be reset (after all, the transformation itself isn't tied to the input table). What do you think?

P-bibs:
I actually think this behavior might also make more sense than the alternative. Since for most transformations the user has to select a context to save the transformation (since a context has to be selected to use the attribute and collection dropdowns), doesn't it stand to reason that they may want to apply the transformation to this context once saved? At the very least, I don't think it hurts to leave the context selected. Maybe we should ask sk.

thomascastleman:
That's fair, maybe we should discuss with sk. This is good for now, I think we can go ahead and merge this.

thomascastleman (approved):

PR #100, opened 2021-06-24
Invalidate cache when attributes get moved

stardust66:
Bug reported in messenger. Cases cache causes issues when attributes get moved, because that does not notify case updates for each individual case, so cached cases are not invalidated properly. This PR invalidates all cases in a context when attributes get moved or deleted.


thomascastleman (approved):
Looks good

P-bibs (approved):
:+1:

PR #101, opened 2021-06-25
Use existing DI creation endpoint

P-bibs:
This PR changes over CODAP-phone to use the existing endpoint for creating data interactive instead of the one we added. This allows us to remove the endpoint we added.


stardust66 (approved):
Looks good.

PR #102, opened 2021-06-25
Further architectural changes

P-bibs:
This PR has a few follow-on changes that are enabled by the data-driven transformations refactor. They  involve simplifying the codebase and de-duplicating information so there's fewer places to update when we make changes.

In general, the end goal is that any data that should have a single source of truth is located in the `transformationList.ts` file. This include types like `BaseTransformationName` (the names of all base transformations) and `TransformationGroup` (all group names in the transformation dropdown), as well as the configuration data for each transformation.

## Changelog
Each change is about one commit message, and the commit messages should approximately match up with the following headers.

### Move transformation grouping data to `transformationList.ts`

Before, we had the grouping data for the transformation dropdown in `Transformation.tsx`. This required duplicating the list of transformation names. Now, that list has been removed and grouping data has been added to each transformation in `transformationList.ts`. `Transformation.tsx` computes the grouping data the first time it renders (using `useMemo`) and then functions identically to before.

### Simplify transformation save button

This refactor removes the `SaveTransformationContext` provider that we were using to allow the `TransformationSaveButton` component to spin off new data interactives. Now, all logic for making new transformations is contained within `TransformationSaveButton`.

This change also removes the hard-coded list of pre-saved transformations in `Transformation.tsx` and replaces it with a computed version that will stay in sync with `transformationList.ts`.

### Simplify render logic in `Transformation.tsx`

Turns two ternaries into one `if` statement.

### Rename `PolymorphicComponent` to `TransformationRenderer`

`PolymorphicComponent` didn't make much sense from the start.


P-bibs:
@thomascastleman @stardust66 feel free to take your time getting to this. It's just my attempt to clean up some of the inconsistencies we've accrued as we change things around. Also, feel free to add on to this branch if there's other things that have been bugging you that you want to fix.

thomascastleman:
This is great, definitely makes things more centralized.

While we're at renaming things, can we find a new name for the `Transformation` component? It feels weird to call it a transformation since it itself is not actually a transformation, it either houses the UI for picking a transformation or houses a saved transformation. I'd even be okay with something like `Plugin` since that fits both cases. Thoughts? @P-bibs @stardust66 

Also, on a related note I wonder if the change to "transformers" means we should shift all uses of "transformation" to "transformer", i.e. "Select a Transformer".

P-bibs:
Agree about the renaming, although I'm not quite sure what a good name would be yet. I think `Plugin` might be too vague.

Also like the change to transformers. Should this be a UI change only, or should we also rename types and variables and such?

thomascastleman:
> Should this be a UI change only, or should we also rename types and variables and such?

For clarity, I think it should probably made to all parts of the code.

Yeah, `Plugin` is vague but also that component is serving two rather disparate purposes. I'm not quite sure what to call it.

thomascastleman:
Just renamed "transformation" to "transformer" in relevant places. Let me know if anything appears to be in error.

Any thoughts on the `Transformer.tsx` name?

thomascastleman:
I renamed our custom UI elements from `CodapFlowX` to just `X`. I also replaced "CODAP Flow" with "Transformers" where appropriate (barring the package.json / repo name).

thomascastleman (approved):
Looks good!

Issue #98, opened 2021-06-24
Don't show hidden attributes in attribute dropdowns


P-bibs:
This affects multi-attribute selector and single-attribute selector.

PR #105, opened 2021-06-25
Fix attribute selector bugs

### LINKED ISSUE #97: Multi attribute allows duplicate attributes
In most cases, the user shouldn't be allowed to select duplicate attributes using the multi attribute selector. We should adjust the component to disallow duplicates (except maybe if a prop flag is passed in).

P-bibs:
Fixes #97 and #98. Also removes an extraneous UI component in `Select Attributes`.

@stardust66 I had to include hiding and unhiding attributes as mutating operations to make sure they invalidated the cache. Can you make sure I did everything right? I'm no super familiar with the caching code.

### DIFF
--- a/src/transformation-components/transformationList.ts
+++ b/src/transformation-components/transformationList.ts
@@ -166,9 +166,6 @@ const transformationList: Record<
       context1: {
         title: "Table to Select Attributes From",
       },
-      textInput1: {
-        title: "Name of New Attribute",
-      },
       dropdown1: {
         title: "Mode",
         options: [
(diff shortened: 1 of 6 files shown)


thomascastleman:
If I select all possible attributes from a dataset and then hide one of them, the dropdown formerly holding the now-hidden attribute goes blank:

missing-attr

Which isn't actually terrible since the transformation goes through correctly anyway (it only groups by State, sample and Year). But it would be nice if we could eliminate this blank dropdown altogether.

stardust66:
@P-bibs, caching looks fine.

I've got a fix for the blank dropdown, should I push it?

P-bibs:
@stardust66 yep go for it

stardust66:
Should be fixed now.

stardust66 (approved):

thomascastleman (approved):
Nice!

PR #106, opened 2021-06-25
gracefully error when not embedded in CODAP

### LINKED ISSUE #99: plugin crashes when not embedded in CODAP

thomascastleman:
This checks if the call to `initPhone` in `App.tsx` fails, and if it does, renders an error message. It also adds a loading screen which is rendered until the call to `initPhone` returns, so that we don't expose the UI to the user when it's not actually embedded in CODAP (this causes all sorts of problems). This loading screen is barely noticeable when the plugin is operating inside CODAP.

Fixes #99.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,57 @@
-import React, { ReactElement } from "react";
+import React, { ReactElement, useState } from "react";
 import Transformation from "./Transformation";
 import { SavedTransformation } from "./transformation-components/types";
 import { initPhone } from "./utils/codapPhone";
+import "./App.css";
 
-// This should be a pure function (ie: only render once)
 export const App = (): ReactElement => {
-  const parsedUrl = new URL(window.location.href);
+  const [loading, setLoading] = useState<boolean>(true);
+  const loadingMsg = <p className="loading">Connecting to CODAP...</p>;
 
+  const [initError, setInitError] = useState<boolean>(false);
+  const initErrorMsg = (
+    <p className="initError">
+      Could not connect to CODAP. Please make sure you are using CODAP Flow
+      within CODAP.
+    </p>
+  );
+
+  /**
+   * Initializes a new data interactive (plugin) with the given name,
+   * and turns off loading once it has succeeded, or indicates an
+   * error should be displayed if it fails.
+   */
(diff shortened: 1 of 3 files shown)


stardust66 (approved):

thomascastleman:
Error when failing to connect:

error

src/App.tsx
```diff
@@ -1,24 +1,57 @@
...
 import { initPhone } from "./utils/codapPhone";
+import "./App.css";
 
-// This should be a pure function (ie: only render once)
 export const App = (): ReactElement => {
-  const parsedUrl = new URL(window.location.href);
+  const [loading, setLoading] = useState<boolean>(true);
+  const loadingMsg = <p className="loading">Connecting to CODAP...</p>;
+
+  const [initError, setInitError] = useState<boolean>(false);
```
P-bibs, line 11: Instead of having two pieces of state that encode the loading/ready status, could we just have something like
```js
type ReadyStatus = "loading" | "ready" | "error";
const [status, setStatus] = useState<ReadyStatus>("loading");
```
That way we don't accidentally make changes in the future that result in and invalid set of state like `loading=false` and `initError = true`.

src/App.tsx
```diff
@@ -1,24 +1,57 @@
...
+  };
 
+  const parsedUrl = new URL(window.location.href);
   const transformation = parsedUrl.searchParams.get("transform");
+  let pluginContent: JSX.Element;
 
   if (transformation === null) {
-    initPhone("CODAP Flow");
-    return <Transformation />;
+    initWithPluginName("CODAP Flow");
```
P-bibs, line 39: Should we have a file for constants like the plugin name? We can also change this to "Transformers"

src/App.tsx
thomascastleman: Nice, this is much slicker. Fixed

src/App.tsx
thomascastleman: Probably a good idea--I have made this change to "Transformers" though on the `further-architecture-changes` branch, so maybe we should have that be there.

P-bibs (approved):
Everything looks good to me now. Feel free to merge.

PR #108, opened 2021-06-28
fix difference from, name changes

thomascastleman:
This changes difference from to match the [corresponding Pyret function](https://www.pyret.org/docs/latest/tables.html#%28part._tables_difference-from%29), and updates the generated dataset descriptions (for this and plain difference) to better explain what they do.

I also did the following name changes:
- "Copy Schema" --> "Copy Structure" since this seems like a more approachable/understandable term
- In several places in the UI where we mention tables, I changed this to "dataset" to match the others we already changed (dropdowns).

## To do
Pick new name for "Running Aggregators"

### DIFF
--- a/src/transformer-components/transformerList.ts
+++ b/src/transformer-components/transformerList.ts
@@ -81,7 +81,7 @@ const transformerList: TransformerList = {
     componentData: {
       init: {
         context1: {
-          title: "Table to Partition",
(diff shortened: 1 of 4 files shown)


stardust66 (approved):

P-bibs (approved):
:+1:

PR #111, opened 2021-06-28
Speed up updates

stardust66:
This makes update listeners execute sequentially, which actually speeds up data fetching. Previously, since the update functions fire off data fetching requests before any data is returned, the same data is fetched multiple times, not making use of our cache. Now, each data fetch waits for the previous one to finish, which means if the same piece of data is fetched multiple times, it will be in the cache by the second fetch.


thomascastleman (approved):
Looks good

P-bibs (approved):

PR #112, opened 2021-06-29
Add ability to save plugin state to CODAP memory

P-bibs:
The following state will now be saved to the CODAP document, and should be available if you close and reopen the file after saving.
* Transformer selector dropdown
* Dropdowns and text areas/inputs within a transformer
* name and purpose statement

To observer this behavior, open a new file with the transformers plugin and save it to your drive. You should notice that interacting with the transformers window dirties the file ("unsaved" will appear in the top left). After a moment the file will save automatically. If you reload now, the plugin window should keep its state.


P-bibs:
Right now, every component that needs to save data uses an arbitrary key on the `saveState` object to store its data, which opens us up to the possibility of collisions if we're not careful. Should we have a centralized schema for the saved state somewhere?

thomascastleman:
It doesn't seem to be saving my selected attributes when I do a Group By. The type (group by) and input table I selected stay there, but the multi-attribute selector is cleared. Any ideas why?

P-bibs:
This is actually caused by a bigger underlying issue related to me forgetting to use `useEffect`. Let me look for a fix and I'll get back to you.

P-bibs:
Should be patched now. Feel free to take another look

thomascastleman:
Nice, seems to work great.

Can you explain more about the collision issue you mentioned? How are components generating keys at the moment?

P-bibs:
The save state that CODAP gives back to us when we load is just an object that we're free to fill out with whatever fields we like. Right now there isn't much discipline with how different components save to those fields. For example, all the state related to dropdown and text box values is saved under a key called `DDTransformation` and the saved function name and purpose statement text boxes are under the keys `currentName` and `description`. If we weren't careful, we could accidentally have one component save its data under a key already used by another component, and erase the previous components data.

thomascastleman:
You mean like if we add an extra UI component and accidentally have it save under "attribute1" and clobber what's already there? I'm not super worried about that--we have to maintain unique names in the `DDTransformerState` type anyway, and it seems like you're using those as storage keys, so I think it's doubtful we'd make that mistake.

thomascastleman (approved):

PR #114, opened 2021-06-29
Fix pivot wider

### LINKED ISSUE #103: Pivot Wider doesn't work with data driven transformations
Pivot wider requires two attribute selectors that reference the same context, but this isn't allowed with our data-driven transformation component.

P-bibs:
Fixes #103 by allowing the attribute/collection selectors in a data driven transformation to target a custom context dropdown.


thomascastleman (approved):
LGTM

stardust66 (approved):

PR #115, opened 2021-06-29
add user-facing documentation to transformers

thomascastleman:
This adds descriptions of all transformers as part of the transformer data specified in `transformerList.ts`.  This info includes a summary of functionality, as well as description of the inputs/outputs of the transformer.

This info is displayed by clicking an info icon that appears once you've selected a transformer:

info

It also appears on saved transformers, which is handy if you want to know more about the underlying transformer that the saved one was based on.

Please review the descriptions themselves and let me know if you have any changes you want to make. Some of them I had a hard time getting to not be verbose and I also want to make sure they're understandable for our target audience (which maybe comes down to eventual user testing).


src/transformer-components/transformerList.ts
```diff
@@ -357,6 +566,13 @@ const transformerList: TransformerList = {
         },
       },
       transformerFunction: { kind: "datasetCreator", func: copy },
+      info: {
+        summary:
+          "Copy produces an duplicate of the given dataset, \
```
stardust66, line 571: Should be "a"

src/transformer-components/transformerList.ts
```diff
@@ -173,6 +233,25 @@ const transformerList: TransformerList = {
         },
       },
       transformerFunction: { kind: "datasetCreator", func: join },
+      info: {
+        summary:
+          "Join combines two datasets into one, based on the values of an \
+        attribute that are shared between the datasets. The output is a copy of \
```
stardust66, line 239: Should be "is shared"

stardust66 (commented):
Looks good! User testing is definitely the best way to decide what kind of language we go with in the end. That said, I think it would be better to not start with the name of the transformation in the summary sections. For example, in Copy, it says "Copy produces a duplicate." We're forced to say "duplicate" since we've already used "copy," but I think it would be better to just say "Produce a copy ..." The transformation names already have the best wording.

P-bibs:
Would it be worth pulling in a UI library to have the documentation be contained in a popup like [this](https://material-ui.com/components/popover/)? That way the main transformation UI doesn't move around as you open and close the documentation.

thomascastleman:
How does this look? @stardust66 @P-bibs 

new-info

stardust66:
Nice! I like it. Though maybe the info icon should go on the right of the drop down? What do you think?

thomascastleman:
I agree, though I'm not sure how to do it since `TransformerREPLView` is what renders the dropdown, and then a `TransformerRenderer` renders the `DataDrivenTransformer` which renders the icon.

P-bibs (approved):
All the descriptions look good to me. I spent a bit trying to figure out how we could make pivot longer/wider make more sense, but it occurs to me that anyone using those transformers will probably be following a curriculum that explains things better than we could ever hope to with a little tooltip, so I think as long as the descriptions are a reasonably good reference we're in the clear.

P-bibs:
> Nice! I like it. Though maybe the info icon should go on the right of the drop down? What do you think?

I agree that this would be the best case scenario but if it can't happen I think we're still alright.

PR #116, opened 2021-06-29
Minor style edits

stardust66:
### Fix extra spacing around submit button for groupby
groupby-before group-by-after

### Fix lack of spacing around editor box
filter-before filter-after

Instead of using `h3` for spacing, we use divs with the class `input-group`, so that elements can be spaced even when there's no title.


P-bibs (approved):

thomascastleman (approved):
Looks good

PR #117, opened 2021-06-29
Fix attribute copying

### LINKED ISSUE #109: Attribute type, unit, precision, not copied correctly by the Copy transformer
To reproduce, take the NOAA Weather dataset and copy it. Notice all the units in the attribute names are gone in the copy, and if you open the attribute info panel, the precision and type information are gone as well.

stardust66:
Fixes #109. Makes `CodapAttribute` type less restrictive. Now, attributes do not need to be numeric to have the `precision` or `unit` properties, which better reflects what the CODAP UI allows. In addition, the type can also be "qualitative", "date", etc., which are options in the type dropdown in the CODAP UI. These options were not in the API documentation, but do have an effect on the UI. For example, the "qualitative" type turns attributes into horizontal bars, like

image

so we should respect them.

### DIFF
--- a/src/utils/codapPhone/index.ts
+++ b/src/utils/codapPhone/index.ts
@@ -372,7 +372,6 @@ export function getDataContext(contextName: string): Promise<DataContext> {
       },
       (response: GetContextResponse) => {
         if (response.success) {
-          console.log(response.values);
           const context = normalizeDataContext(response.values);
           Cache.setContext(contextName, context);
           resolve(context);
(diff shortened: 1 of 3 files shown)


src/utils/codapPhone/index.ts
```diff
@@ -370,6 +372,7 @@ export function getDataContext(contextName: string): Promise<DataContext> {
       },
       (response: GetContextResponse) => {
         if (response.success) {
+          console.log(response.values);
```
thomascastleman, line 375: Is keeping this log intended?

stardust66: Good catch. I'll get rid of it.

thomascastleman (approved):
Nice, this cleans things up and solves this issue.

P-bibs (approved):

PR #118, opened 2021-06-30
ensure all transformations avoid name collisions

thomascastleman:
This goes back and patches some uses of generated names that weren't ensuring uniqueness. 

I also updated the centralized unique naming function to use `{n}` instead of `(n)` (not ideal, but `(n)` is interpreted as specifying a unit and so won't end up part of the attribute name. And it doesn't like `[n]` for data contexts for some reason).


stardust66 (approved):
Looks good!

P-bibs (approved):

PR #120, opened 2021-07-01
Add undo functionality for transformer application

P-bibs:
This PR adds the ability to use the CODAP undo button to delete tables that have just been created by a transformer application. At present, this is the only undoable functionality.

## Design decisions
If the user takes certain actions that are hard to undo/redo (like creating/deleting tables), CODAP clear it's undo/redo stack and start over (the undo/redo buttons will be grayed out). CODAP informs us of this so we can also clear the plugins internal undo/redo stack that we're tracking.

Since all our undo functionality involves deleting tables, the user will only ever be able to undo one transformer application before CODAP clears its undo functionality. This PR implements the following hack to get around that: if CODAP tells us to clear the undo stack but we still have undoable actions left, instead of clearing the undo stack we call `notifyUndoableActionPerformed()` once for every remaining action in the plugin's undo stack. This tells CODAP to put the plugin back in the undo stack and lets the user perform more "undo"s of transformer application.

However, this could result in undesirable behavior. Suppose we have the following stack of events:
```
<oldest event>
Transformers: Apply build column transformation to create table "A"
CODAP: rename attribute in "A"
CODAP: move attribute in "A"
CODAP: add case to "A"
CODAP: add case to "A"
Transformers: Apply copy transformation to "A"
<newest event>
```
With the current functionality, if the user presses undo once, the copy of "A" will be deleted. If they press undo again, "A" will be deleted. This might be shocking, since the user probably expects to undo adding the case to "A".

**I think this is a decision we have to make regarding the risk of having more powerful behavior at the of confusing edge cases. I'd be curious to hear your thoughts.**

## Troubles with redo
It looks like we won't be able to have redo without some hefty CODAP changes. Here's why:

Suppose we tell CODAP an undoable action has been performed when someone hits "apply transformer" to create a table. Then, if the user presses undo in CODAP, CODAP will inform us and we can delete the table. 

Ideally, the redo button should now be enabled, and if the user hits "redo" CODAP will inform us and we can re-create the deleted table. The issue is, CODAP treats deleting a table as an undo-clearing action. Since our undo functionality relies on deleting a table, the undo/redo stack is cleared and the user can't hit redo to recreate the table. This means even though our architecture _could_ support undo/redo of transformation application, the user won't be able to use them since CODAP keeps clearing the undo/redo stack.

## What else should be undoable
Are there any other plugin-related actions that should be undoable? We already get undo for free in an expression editor from codemirror.

### DIFF
--- a/src/transformer-components/DataDrivenTransformer.tsx
+++ b/src/transformer-components/DataDrivenTransformer.tsx
@@ -304,13 +304,6 @@ const DataDrivenTransformer = (props: DDTransformerProps): ReactElement => {
 
         const textName = await createText(name, String(result));
 
-        // Add action to undo stack
-        pushToUndoStack(
-          `Apply ${base} transformer`,
-          () => deleteText(textName),
-          transform
-        );
-
         // Workaround because the text doesn't show up after creation
         // See https://codap.concord.org/forums/topic/issue-creating-and-updating-text-views-through-data-interactive-api/#post-6483
         updateText(textName, String(result));
--- a/src/transformer-components/DataDrivenTransformer.tsx
+++ b/src/transformer-components/DataDrivenTransformer.tsx
@@ -244,7 +244,7 @@ const DataDrivenTransformer = (props: DDTransformerProps): ReactElement => {
   useEffect(() => {
     async function fetchSavedState() {
       const savedState = (await getInteractiveFrame()).savedState;
-      if (savedState.DDTransformation) {
+      if (savedState && savedState.DDTransformation) {
         setState(savedState.DDTransformation);
       }
     }
--- a/src/transformer-components/DataDrivenTransformer.tsx
+++ b/src/transformer-components/DataDrivenTransformer.tsx
@@ -326,7 +326,7 @@ const DataDrivenTransformer = (props: DDTransformerProps): ReactElement => {
       } else if (typeof result === "object") {
         // This is the case where the transformation returns a dataset
         const newContextName = await applyNewDataSet(result, name, description);
-        pushToUndoStack("Transformation applied", () =>
+        pushToUndoStack(`Undo ${base} transformer`, () =>
           deleteDataContext(newContextName)
         );
         if (order.includes("context1") && state["context1"] !== null) {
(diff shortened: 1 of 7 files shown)


thomascastleman:
Yep, I definitely see what you mean with edge cases. I think this is something we should demo for sk and see what he thinks--it certainly is powerful to have any level of undo.

As for other features using undo, I'm not sure. I don't think we need to be able to undo creation of a saved transformation (you can just delete it, + with the editing functionality there won't be much need). 

Maybe this should apply to the textbox output of Average / Dot Product? Seems like any output we generate should be retractable if possible.

P-bibs:
Yep, text boxes were an oversight and I'm adding them in now. One issue however: if the user hits undo after applying a transformer that creates a text box, then we delete the text box. However, CODAP treats deleting that text box as an action, so the user can then hit undo again to get the text box back. I'm not really sure what the desired behavior in this situation is. Maybe we should add a CODAP endpoint to clear the undo stack so that we can avoid this situation?

P-bibs:
Also, I'm starting to find more situations that make the extended undo  mechanism described in the original PR description undesirable. Here's an example:

* Make the microdata table
* Apply the copy transformer to it
* Change one of the cases in the original table and watch it change in the copy
* Hit undo, expecting the change you made to the original table to be undone. Instead, the copied table gets deleted.

P-bibs:
I talked with Jonathan and we decided on only having one level of undo to avoid confusing behavior. Additionally, the plugin doesn't need to handle undo/redo for text boxes since once a CODAP bug is fixed CODAP will handle those automatically.

stardust66 (approved):

src/transformer-components/DataDrivenTransformer.tsx
```diff
@@ -4,7 +4,9 @@ import {
   createText,
   getInteractiveFrame,
   notifyInteractiveFrameIsDirty,
+  deleteDataContext,
   updateText,
+  deleteText,
```
thomascastleman, line 9: Looks like this `deleteText` import isn't used

src/utils/codapPhone/index.ts
```diff
@@ -34,6 +34,10 @@ import {
   removeContextUpdateListenersForContext,
   removeListenersWithDependency,
   callAllInteractiveStateRequestListeners,
+  popFromUndoStackAndExecute,
+  popFromRedoStackAndExecute,
+  undoStack,
```
thomascastleman, line 39: `undoStack` is also not used

thomascastleman (approved):
Looks good! Just a few lints that can be fixed it seems. I like one level of undo--I think it covers a good amount of use cases for undo and is definitely better than not having it at all.

PR #121, opened 2021-07-02
small UI improvements

thomascastleman:
- Fix font in purpose statement textbox to match rest of UI
- Add proper errors to the save transformer logic
- Make default plugin size larger
- Only render "more info" icon in transformer REPL

### DIFF
--- a/src/transformer-components/DataDrivenTransformer.tsx
+++ b/src/transformer-components/DataDrivenTransformer.tsx
@@ -322,23 +322,28 @@ const DataDrivenTransformer = (props: DDTransformerProps): ReactElement => {
 
   return (
     <>
-      <Popover
-        icon={<InfoIcon htmlColor="#72bfca" fontSize="small" />}
-        tooltip={`More Info on ${base}`}
-        innerContent={
-          <>
-            <p>{splitIntoParagraphs(info.summary)}</p>
-            <p>
-              <b>Inputs: </b>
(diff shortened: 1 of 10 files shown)


src/transformer-components/DataDrivenTransformer.tsx
```diff
@@ -359,23 +361,28 @@ const DataDrivenTransformer = (props: DDTransformerProps): ReactElement => {
...
-              <b>Produces: </b>
-              {splitIntoParagraphs(info.produces)}
-            </p>
-          </>
-        }
-      />
+      {/* Only render info icon if NOT a saved transformation. */}
+      {saveData ? (
+        <></>
+      ) : (
```
P-bibs, line 367: I'm pretty sure the following is idiomatic in React:
```diff
- {condition ? <Element /> : <></>}
+ {condition && <Element />}
```
So maybe this should be 
```jsx
{!saveData && <Popover ... />
```
Although I think the version you currently have is more readable. I'm not sure which should win out, convention or readability. @stardust66 thoughts?

P-bibs (approved):

src/transformer-components/DataDrivenTransformer.tsx
thomascastleman: I'm okay with either though I see why the latter is conventional, it seems to leave out a lot of clutter.

PR #122, opened 2021-07-02
allow editing of saved transformers

thomascastleman:
This adds support for editing transformers that have already been saved. You are able to edit any of the inputs to the transformer, but not its name/purpose statement.

You can edit by clicking the Edit button (seen below), at which point disabled inputs will become enabled. The button toggles whether or not inputs are disabled.

edit

### DIFF
--- a/src/transformer-components/SavedTransformerView.tsx
+++ b/src/transformer-components/SavedTransformerView.tsx
@@ -15,6 +15,8 @@ import {
   addInteractiveStateRequestListener,
   removeInteractiveStateRequestListener,
 } from "../utils/codapPhone/listeners";
+import "../ui-components/TransformerSaveButton.css";
+import "./SavedTransformerView.css";
 
 /**
  * SavedTransformerView wraps a saved transformer in other important info
@@ -28,6 +30,7 @@ function SavedTransformerView({
   const [errMsg, setErrMsg] = useState<string | null>(null);
   const [editable, setEditable] = useState<boolean>(false);
   const [savedTransformer, setSavedTransformer] = useState(urlTransformer);
+  const [saveErr, setSaveErr] = useState<string | null>(null);
 
   // Load saved state from CODAP memory
   useEffect(() => {
@@ -68,13 +71,20 @@ function SavedTransformerView({
 
   return (
     <div className="transformer-view">
+      {editable && (
+        <div className="editing-indicator">
+          <p>You are editing this transformer...</p>
+        </div>
+      )}
       {editable ? (
         <TextInput
           value={savedTransformer.name}
-          onChange={(e) =>
-            setSavedTransformer({ ...savedTransformer, name: e.target.value })
-          }
+          onChange={(e) => {
+            setSavedTransformer({ ...savedTransformer, name: e.target.value });
+            setSaveErr(null);
+          }}
           placeholder={"Transformer Name"}
+          className="saved-transformer-name"
           onBlur={notifyStateIsDirty}
         />
       ) : (
@@ -86,13 +96,15 @@ function SavedTransformerView({
       {editable ? (
         <TextArea
           value={savedTransformer.description || ""}
-          onChange={(e) =>
+          onChange={(e) => {
             setSavedTransformer({
               ...savedTransformer,
               description: e.target.value,
-            })
-          }
+            });
+            setSaveErr(null);
+          }}
           placeholder="Purpose Statement"
+          className="purpose-statement"
           onBlur={notifyStateIsDirty}
         />
       ) : (
@@ -106,7 +118,15 @@ function SavedTransformerView({
       />
       <button
         id="edit-button"
-        onClick={() => setEditable(!editable)}
+        onClick={() => {
+          // if going to non-editable (saving) and name is blank
+          if (editable && savedTransformer.name.trim() === "") {
+            setSaveErr("Please choose a name for the transformer");
+            return;
+          }
+
+          setEditable(!editable);
+        }}
         title={
           editable
             ? "Save changes made to this transformer"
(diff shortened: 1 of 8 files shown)


thomascastleman:
I want to extend this to allow editing of saved name / purpose statement, but I also realized editing introduces weird interactions between our URL parameterization of state + the saved state of the document. 

Right now if I save a transformer, edit it, then reload that frame, my edits are erased (because they were just part of the UI and when the reload happens all state is recovered from the URL which was created when I originally created the saved transformer). 

However, if I save, make edits, then _save the document_, my edits are preserved--but the saved URL of the saved transformer now conflicts with the saved document state.

Is there some way we could not have the saved state being tracked in two separate places (the URL and the saved state of the document)? Thoughts @P-bibs ?

P-bibs:
Yeah I had been mulling this over since Bill brought it up on the first call where we talked about state. I think what you have going actually makes sense and is a nice side effect of the way our current architecture is set up. The URL params initialize the transformation, but since the loading from saved state happens _after_ the URL params are parsed, the saved state takes priority. This definitely does feel a little sloppy since the two sourced conflict, but there's not really anyway around it. The only way we can initialize plugins with state is via URL params, and the only way we can carry state through page reloads is saving to CODAP. 

Critically, CODAP doesn't let us modify the plugin URL after plugin creation. If that were to change, I suppose once the saved transformer is initialized, we could wipe the URL params so there aren't two sources of state.

thomascastleman:
Good points, that does make sense. I suppose it's also a pretty unlikely case if a user makes edits to a saved transformer, then reloads that iframe on its own. If they reload the whole page without saving the document, they should expect to see their edits disappear anyway.

I wonder now if students will confuse the notion of saving a transformer with saving the CODAP document itself. Maybe this is something to discuss with sk / Emmanuel but we might want to stay away from "save" lingo to avoid that? (perhaps it's just a "transformer definition" in keeping with our definitions/interactions metaphor). Maybe it's not an issue.

I will work on getting edit capabilities for the name/purpose statement.

thomascastleman:
Really weird thing that is happening--when I focus on a disabled expression editor and type into it, nothing happens (as expected). But when I later enable it for editing, the text I typed while it was disabled appears in the box (as though I had typed it). @stardust66 Any idea why this is happening?

stardust66:
That's really weird. I don't really know off the top of my head but let me reproduce it.

stardust66:
@thomascastleman. Found a fix. Should I push to this branch? I guess this bug doesn't really show up otherwise.

thomascastleman:
Yep that'd be great thanks

P-bibs:
Good catch on the "save" terminology I definitely think that's something the BS gang will pick up on and hopefully have suggestions.

thomascastleman:
@stardust66 @P-bibs Let me know what you think. You can now edit name/purpose statement, as well as the inputs to the transformer.

src/transformer-components/SavedTransformerView.tsx
```diff
@@ -15,19 +29,121 @@ function SavedTransformerView({
...
   return (
     <div className="transformer-view">
-      <h2>
-        {urlTransformer.name}
-        <span id="transformerBase"> ({urlTransformer.content.base})</span>
-      </h2>
-      {urlTransformer.description && <p>{urlTransformer.description}</p>}
+      {editable && (
+        <div className="editing-indicator">
+          <p>You are editing this transformer...</p>
```
P-bibs, line 77: We can probably leave off the `...`

src/transformer-components/SavedTransformerView.tsx
```diff
@@ -15,19 +29,121 @@ function SavedTransformerView({
...
+            });
+            setSaveErr(null);
+          }}
+          placeholder="Purpose Statement"
+          className="purpose-statement"
+          onBlur={notifyStateIsDirty}
+        />
+      ) : (
+        savedTransformer.description && <p>{savedTransformer.description}</p>
+      )}
```
P-bibs, line 113: Can we add an `<hr/>` between the name/purpose statement and the actual transformer data so when you're editing a saved transformation it's clear which boxes are the transformation and which are name/purpose statement.

P-bibs (approved):
One question that comes to mind: if I edit a saved transformer after already making a new table with it, does that table change to reflect the new transformer? Certainly simpler to say :"no", but then we might end up with confusing situation where a table indicates it was made from the transformer `is-old` but that transformer has since changed definition.

src/transformer-components/SavedTransformerView.tsx
thomascastleman: Always or just while you're editing?

thomascastleman:
That's a good point, though it is worth noting that (at least right now), the output dataset(s) has no indication of a saved transformer by which it was created. So, in the case that a saved transformer is updated, the dataset description of that previous output would remain correct--it describes the specific parameters that were used to create that dataset. 

If we do do something like a flow graph and end up mentioning particular saved transformations in it (i.e. as labels on the arrows for instance), then I think this may become more of an issue.

P-bibs:
Ahh that's a good point. Although I think it raises the question of if table titles should reflect the base transformer that was used to make them or the name of the saved transformer. For instance if I save a filter transformer called "is-old" and then apply it to the People table, should the output be "Filter of (People)" or "is-old of (People)" (or maybe "is-old applied to (People)").

src/transformer-components/SavedTransformerView.tsx
P-bibs: My instinct is just while editing as the styling we use when not editing is sufficient to distinguish the two sections, but if it looks better to have it always then that works to.

thomascastleman:
Yeah that's a good question--I could see either being used and I really don't have an intuition of which would be better for the Bootstrap curriculum, so maybe it's worth asking about.

@stardust66 Does this look good to you to merge?

stardust66 (approved):
Yup. Looks good.

PR #123, opened 2021-07-05
Don't resize plugin to default dimensions when opening existing file

### LINKED ISSUE #110: Transformers window is resized to be too small when opening a file that contains the transformers plugin

P-bibs:
Fixes #110. When opening a new file with `Transformers`, it should still be resized to the default dimensions. However, if you open a CODAP file that contains `Transformers`, it will stay at its current size.

Also added a useEffect hook to stop initialization logic in `App` from being fired multiple times.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,54 +4,57 @@ import { SavedTransformer } from "./transformer-components/types";
 import TransformerREPLView from "./transformer-components/TransformerREPLView";
 import { initPhone } from "./utils/codapPhone";
 import "./App.css";
+import { useEffect } from "react";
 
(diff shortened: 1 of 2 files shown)


thomascastleman:
Hmm, it seems to be initialized to the saved dimensions for an instant and then gets immediately resized back to default for me.

P-bibs:
Is this with the REPL or a saved transformation?

thomascastleman:
Hmm, nevermind, it seems to work now for both.

thomascastleman (approved):

stardust66 (approved):

PR #127, opened 2021-07-06
recognize color values when formatting

thomascastleman:
This adds support for recognizing CODAP color values when we go to format colors for printing. For instance, this is used in the titles of partitioned datasets:

partition

It also comes up in error messages:

color-error

I figure it's good to expose the underlying hex/rgb/rgba color values--in the event the user didn't create it themselves and doesn't understand it, they'll still see "color" and know it was a color, and in the event they did create it of their own accord, knowing the color values might provide critical info (I prefer this to just showing "a color").

### DIFF
--- a/src/transformers/util.ts
+++ b/src/transformers/util.ts
@@ -182,8 +182,7 @@ export function pluralSuffix<T>(word: string, describing: T[]): string {
 }
 
 /**
- * Converts a CODAP cell value into a user-friendly string
- * for printing in error messages.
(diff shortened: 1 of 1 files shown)


P-bibs:
This looks great, especially considering I'm working on tweaking numeric compare so we get more color hues.

P-bibs (approved):

stardust66 (approved):
Looks good!

PR #128, opened 2021-07-06
fix partition naming bug and make partitioned values clearer

thomascastleman:
This fixes a bug that was causing the data context name used in partitioned table titles/descriptions to be the actual name and not the title (unlike all the other transformers).

I've also modified the names of output tables in partition so that the values are nearer to the front of the name--however I didn't really find a non-awkward way to have the name start directly with the value. Let me know what you all think of this "{attribute} = {value} in {dataset}" naming convention.

name-changes

### DIFF
--- a/src/transformers/partition.ts
+++ b/src/transformers/partition.ts
@@ -51,9 +51,9 @@ export const partitionOverride = async (
     // return both the datasets and their names
     return partitioned.map((pd) => [
       pd,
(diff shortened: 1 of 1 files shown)


stardust66 (approved):
Nice! I like the new naming convention, and also can't think of a good way to put the values at the very front.

P-bibs (approved):
I think this is great and nicely solves the issue Kathi pointed out yesterday.

Issue #125, opened 2021-07-05
Accessibility issues on numerical compare


P-bibs:
Green and red can have insufficient contrast for certain types of color blindness. Additionally, it would be good if the saturation of the color was proportional to the value of the difference.

PR #129, opened 2021-07-06
Add color gradation to numeric compare

P-bibs:
Now when performing a numeric compare, the saturation of the colors is proportional to the magnitude of the difference. Ideally we could also solve #125 with this PR, but I'm not sure what colors to use besides red and green.

### DIFF
--- a/src/utils/colors.ts
+++ b/src/utils/colors.ts
@@ -7,7 +7,7 @@ export const RED: Color = [255, 0, 0];
  * Given two colors and a value between 0 and 1, calculates the color that
  * lies "between" the `from` and `to` colors.
  *
- * Progress of 0 corresponds to the `from` color and progress of ` corresponds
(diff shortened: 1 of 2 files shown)


src/utils/colors.ts
```diff
@@ -0,0 +1,30 @@
+type Color = [number, number, number];
+export const GREY: Color = [230, 230, 230];
+export const GREEN: Color = [0, 255, 0];
+export const RED: Color = [255, 0, 0];
+
+/**
+ * Given two colors and a value between 0 and 1, calculates the color that
+ * lies "between" the `from` and `to` colors.
+ *
+ * Progress of 0 corresponds to the `from` color and progress of ` corresponds
```
stardust66, line 10: I think you meant to say "progress of 1" here?

stardust66 (approved):

thomascastleman (approved):

PR #138, opened 2021-07-06
tests for core transformer functions

thomascastleman:
This is a PR for adding a test suite for the core unchecked transformer functions. In most cases the unchecked functions are pretty pure, so it should be straightforward to test, however there are some that make requests (for instance to evaluate expressions), which will be trickier to test.

I've set this up so we can have small test datasets in `transformers/__tests__/data.ts`, and reference these in our tests.


thomascastleman:
Note: Some tests are failing because of attribute validation stuff that hasn't been implemented and probably should be done in a separate PR.

thomascastleman:
Note: fixes for the failing partition tests are on my `tc-tests` branch.

PR #140, opened 2021-07-08
Make updates persistent across saves

stardust66:
Instead of using individual listeners to perform flow updates, keep a record of all active transformations and use that to perform updates. This record can then be included in the save state. We can also use it to draw a flow diagram.

Right now, updates for `datasetCreator` transformers are persistent across saves. Still haven't implemented it for the `fullOverride` transformers.


stardust66:
The gist of the PR is that it keeps track of a list of active transformations, and instead of registering callbacks, we add to the list when creating a transformation. Since we do all updates based on the record and the record gets stored in the save state, the updates will persist.

I added context update hooks and context deletion hooks (not sure these are the best names). Instead of choosing which callbacks to call when specific contexts update, the hook is a single function that gets called with the name of the updated context, so it can decide what to do based on that.

To make partition work, I had to factor out the update code into `partitionUpdate`, and store that in `transformerList` so it can be looked up by the update code. The way I'm handling full override transformers is that they get their custom save state that they can refer to in their update functions. The types may have to be reworked if we add a second full override transformer.

Regarding where this code should go, I feel like putting it under `util` is not quite right, but I'm not sure where else to put it.

thomascastleman:
This is great, feels a lot more robust.

One thing: If I change an input table in a way that causes an error while updating, it seems I get the plain error message. It would be nice to have something like we did before, where the message includes "Error updating ____: ..." so we know what happened.

thomascastleman (approved):

stardust66:
Good point. I'll get that back.

P-bibs (approved):
Looks good to me! There's a few log statements that have been left in throughout, are those on purpose?

PR #141, opened 2021-07-09
Parameterize formula evaluater in transformers that use formulas

P-bibs:
This PR modifies all transformers that require evaluating a formula to accept a function as a parameter that implements that evaluation logic.

The existing CODAP phone `evalExpression` function is used as a default parameter, but the `transformers/__tests__/data.ts` file includes a `jsEvalFormula` function that can be passed instead to use the js `eval` function as a mock for the CODAP engine.

The tests that actually use this functionality aren't included here. I have them on a separate branch. However, I've also included some additional testing datasets and some changes to the existing test functions.

### DIFF
--- a/src/transformers/__tests__/data.ts
+++ b/src/transformers/__tests__/data.ts
@@ -30,6 +30,9 @@ function makeRecords(
   valueLists: unknown[][]
 ): Record<string, unknown>[] {
   return valueLists.map((valueList) => {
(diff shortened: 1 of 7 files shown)


thomascastleman (approved):
Nice, I love the "fully featured dataset" and will definitely incorporate that into my tests.

stardust66 (approved):

PR #144, opened 2021-07-13
add median, mode, std deviation

thomascastleman:
This renames "Average" --> "Mean", and adds three new single-value transformers: Median, Mode, and Standard Deviation.

At the moment, all of these error on a dataset with no records, as well as if non-numeric values are given.

Please look over these and let me know if there are any edge cases you think aren't handled correctly.

### DIFF
--- a/src/transformers/util.ts
+++ b/src/transformers/util.ts
@@ -473,3 +473,30 @@ export function shallowCopy<T>(x: T): T {
 }
 
 export const cloneAttribute = shallowCopy;
+
+/**
+ * Extracts a list of numbers from an attribute in a dataset that is expected
+ * to contain numeric values. Errors if the attribute is undefined for some
+ * records, or if the attribute's values parse to non-numeric.
+ *
+ * @param dataset The dataset to extract from.
+ * @param attribute The attribute containing numeric values.
+ * @returns A list of the attributes values parsed to numbers.
+ */
+export function extractNumericValues(
+  dataset: DataSet,
+  attribute: string
+): number[] {
+  return dataset.records.map((record) => {
+    if (record[attribute] === undefined) {
+      throw new Error(`Invalid attribute name: ${attribute}`);
+    }
+    const numericValue = parseFloat(String(record[attribute]));
+    if (isNaN(numericValue)) {
+      throw new Error(
+        `Expected number, instead got ${codapValueToString(record[attribute])}`
+      );
+    }
+    return numericValue;
+  });
+}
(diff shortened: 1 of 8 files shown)


src/transformers/util.ts
```diff
@@ -473,3 +473,30 @@ export function shallowCopy<T>(x: T): T {
...
+/**
+ * Extracts a list of numbers from an attribute in a dataset that is expected
+ * to contain numeric values. Errors if the attribute is undefined for some
+ * records, or if the attribute's values parse to non-numeric.
+ *
+ * @param dataset The dataset to extract from.
+ * @param attribute The attribute containing numeric values.
+ * @returns A list of the attributes values parsed to numbers.
+ */
+export function extractNumericValues(
```
P-bibs, line 486: The name of this function makes me think we're going to extract the numeric values and throw away the rest, as opposed to erroring if there are non-numeric values. Could we rename to something like `ensureNumericValues`/`convertToNumericValues`/`extractAttributeAsNumeric`

src/transformers/util.ts
```diff
@@ -473,3 +473,30 @@ export function shallowCopy<T>(x: T): T {
...
+  dataset: DataSet,
+  attribute: string
+): number[] {
+  return dataset.records.map((record) => {
+    if (record[attribute] === undefined) {
+      throw new Error(`Invalid attribute name: ${attribute}`);
```
P-bibs, line 492: I'm not sure `Invalid attribute name: ${attribute}` is the best error here. I checked with the API tester and it seems like you can have an attribute be undefined for some records and defined for others, so it might make more sense to have an error like `Found case that is missing value for ${attribute}`.

Granted, if a user is encountering this issue it's probably because another plugin messed with the table in ways it shouldn't have, so even if we have a more helpful error message there's not much they can do about it.

thomascastleman: Hmm I didn't realize that was possible. I suppose the correct check here (and we should do this in all the other transformers too) is to see if _every_ `record[attribute]` is undefined, then we can raise the "Invalid attribute" error.

I'm not sure about calling it a missing value though, since that's a well-defined thing in CODAP--what does an attribute being `undefined` for a single case even look like in the UI? Is it blank just like missing values?

P-bibs: Yes, they just appear as blank cells, so to the user they're indistinguishable from the empty string. Could we check whether the attribute exists in any of the collections instead? That might be more fool proof.

thomascastleman: Yes, though I'm inclined to do this on a separate PR, and for all the transformers that take attributes (since they mostly use this method at the moment).

I guess if they are indistinguishable from the user perspective it's okay for us to use that name in the error message.

stardust66 (approved):
Looks good!

PR #145, opened 2021-07-13
Remove structural compare and separate other compares

P-bibs:
This PR removes structural compare and divides numeric compare and categorical compare into two separate transformations. 

Additionally, numeric compare and categorical compare now require both attributes to be form the same dataset. Further, numeric compare has been modified in the following ways:
* all attributes are now kept in the output dataset (instead of just the attributes being compared)
* the collection that the comparison attributes are added to is the rightmost collection of the two collections the selected attributes are in

image


P-bibs:
I ended up making numeric compare not require two attributes in the same collection, so we could combine categorical and numeric compare back into one transformation if we wanted. We can discuss in meeting on Wednesday.

P-bibs:
Also, I'm not raising any error right now if a numeric compare is performed with an attribute that has no numeric values. Do you think we should have error messages on per-case basis (the error message is in the table itself) or should we surface an error in the transformers window if too many of the attribute values are non-numeric?

thomascastleman:
Hmm, I think for consistency with our other transformers, I would prefer surfacing some sort of error in the transformers window for non-numeric values. I have a feeling sk will want this too but we can discuss as well.

I don't have any strong feelings about separate vs. together for numeric/categorical, so we should probably bring that up and ask.

P-bibs:
Yep, after looking at the mean/median/stdev PR I think forcing numeric values is the right way to go. I'll work on a patch.

src/transformer-components/transformerList.ts
```diff
@@ -339,32 +339,27 @@ const transformerList: TransformerList = {
         context1: {
           title: "First Dataset to Compare",
```
thomascastleman, line 340: This should probably just be "Dataset to Compare" or even "Dataset to Compare Within"

thomascastleman:
Now with mean/median/mode/etc all ignoring missing values, should numeric compare have the same behavior? To be honest I'm not sure what should allow missing values and what should let them go silently by...

Erroring on other non-missing non-numeric types seems right for sure, but missing values seem to require special consideration.

stardust66:
I guess ignoring missing values in mean/median/mode just treats the whole row as missing, but what does it mean when a value being compared is missing? What should go in the difference column? I don't think missing values should be interpreted as 0, but maybe it's okay if the difference is left blank?

thomascastleman:
Right yeah my guess would be if missing values were allowed, the Difference and Compare Status columns would both just contain missing values.

stardust66 (approved):
This looks good in general. Besides what to do with missing values, my other concern is that users would forget what is being compared since all the columns are there. They can always check the description but I wonder if there's a more prominent place we could include this information.

thomascastleman (approved):

P-bibs:
I agree that missing values should be skipped. I've added a fix for that. Regarding attribute names: I had the same concern. The UI is a little clunky if you try to fit both attribute names in the new attribute. Should we try and put them in the table title?

Issue #148, opened 2021-07-15
"Invalid attribute" when attribute is valid


thomascastleman:
1. Create a dataset with two collections (e.g. Microdata)
2. Transform this dataset with something that uses an attribute _in the child collection_ (e.g. Transform Column)
3. Delete all cases in the input dataset (this update should flow to the output--all good)
4. Start defining a case in the input by defining something in the parent collection

Now you should get an error during the update, indicating that the attribute (from the child collection) is invalid, even though it clearly is valid. This is what Paul mentioned in #144 -- cases can be `undefined` for a given attribute in weird circumstances and this is one of them.

This raises the question of what we want to do in this case: Raise a different error? On the other hand we might want to make these `undefined` values behave like missing values (that's what they look like to the user, after all), in which a case the Transform Column would go through correctly.

Thoughts? @stardust66 @P-bibs

P-bibs:
I'm partial to the treating them like missing values approach, since having two values that look the same to the user (empty string and `undefined`) but behave different seems like it could lead to frustrating behavior.

thomascastleman:
Since #151, we no longer get an "Invalid attribute" error when the attribute is valid (we've switched to treating those `undefined`s as missing values. So I think it's fair to close this. 

But worth noting that the example given above still errors but now from the CODAP end--we send records with `undefined`, so it considers the attribute "unrecognized". This is arguably a CODAP bug, since `undefined` _appears_ graphically as a missing value, so it should either be treated as such or just not represented using `undefined`.

Good to close? @stardust66 @P-bibs

PR #151, opened 2021-07-16
validate attributes in a uniform way

thomascastleman:
This adds a mechanism for ensuring that when a transformer takes a dataset and an attribute(s) from that dataset, we can check that the attribute exists within the dataset (by searching collections).

Previously, "Invalid attribute" errors were mostly being detected when we were processing a row and found that the given attribute was `undefined` for that row. We later discovered that rows can be `undefined` for an attribute that is indeed part of the dataset (see #144 and #148). Relying on row processing to detect these errors was brittle because 1) if there are no rows, attribute errors wouldn't be detected, and 2) the above situation, in which an attribute is actually valid but `undefined` in rows.

This checks attributes separately from handling of undefined values in rows. This PR also moves to treating `row[attr] === undefined` as a _missing value_, instead of an error--this is because when this happens, it appears exactly as a missing value does to the user.

### DIFF
--- a/src/transformers/count.ts
+++ b/src/transformers/count.ts
@@ -1,6 +1,6 @@
 import { DataSet, TransformationOutput } from "./types";
 import { CodapAttribute, Collection } from "../utils/codapPhone/types";
-import { listAsString, eraseFormulas, shallowCopy, pluralSuffix } from "./util";
+import { listAsString, eraseFormulas, shallowCopy, pluralSuffix, validateAttribute } from "./util";
 import { uniqueName } from "../utils/names";
 import { DDTransformerState } from "../transformer-components/DataDrivenTransformer";
 import { getContextAndDataSet } from "../utils/codapPhone";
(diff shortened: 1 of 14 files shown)


thomascastleman:
Worth noting: this doesn't actually fix #148 -- it just changes the error you get. It now errors when it goes to the formula evaluator--it sees the child attribute as "unrecognized" and this is likely because it's undefined. I think the ideal behavior would be to treat the not yet defined child attribute as a missing value (as it looks), but this might involve ensuring that when we send a formula off for evaluation, all valid attributes are bound (whether to a real value or `""` for missing values).

P-bibs:
Just to clarify: are you suggesting that before we send a formula off to be evaluated on a table, we call a function to go over that table and replace any `undefined`s with `""`s? I think this could work, and hopefully wouldn't result in too much of a performance hit.

thomascastleman:
Yes, something like that. In effect, this would normalize missing values to `""`, even though they can appear as both `undefined` and `""`.

thomascastleman:
This look okay? @stardust66 @P-bibs It will fix several broken tests so I'm eager to merge if possible.

P-bibs (approved):
Nice work! I don't think I had realized how many different places we had duplicated this functionality.

Issue #21, opened 2021-05-18
generated table attributes not draggable at first


thomascastleman:
After creating a table through a transformation, the table's attributes cannot be dragged around to reorder them, until another transformation takes place, at which point they become draggable (but not for the latest transformed table).

thomascastleman:
After further investigating, I am pretty sure this is a bug on the CODAP end (it happens even with simple API requests to create data contexts, outside any involvement of our transformations/updating/etc).

I have reported it [here](https://codap.concord.org/forums/topic/attributes-not-draggable-on-most-recent-api-created-data-context/).

thomascastleman:
Fixed [here](https://github.com/concord-consortium/codap/commit/df70019d2c9bff32bcb5dbd107bbd62872489d66).

PR #152, opened 2021-07-16
Make transformation output tables immutable and add mutable clone operation

stardust66:


src/transformers/mutableClone.ts
```diff
@@ -0,0 +1,28 @@
...
+    return;
+  }
+
+  const { context, dataset } = await getContextAndDataSet(inputDataCtxt);
+  const ctxtName = readableName(context);
+
+  applyNewDataSet(
+    uncheckedCopy(dataset),
+    `Clone of ${ctxtName}`,
+    `A copy of the ${ctxtName} dataset.`
```
thomascastleman, line 26: I think the dataset description here should indicate the special properties of this transformer: that it is a copy, it is mutable, and it won't update when the input does.

src/transformers/util.ts
```diff
@@ -508,3 +508,16 @@ export function extractAttributeAsNumeric(
...
+  for (const c of newCollections) {
+    if (c.attrs) {
+      c.attrs.map((attr) => (attr.editable = false));
+    }
+  }
+  return {
+    collections: newCollections,
+    records: dataset.records,
+  };
+}
```
thomascastleman, line 523: Nice, this is already a big improvement over the previous. I wonder, is there any way to disable dragging (or renaming) of attributes? Or is editable=false all we have for immutability (maybe this is something we could request from Bill and Jonathan--it might not be hard to do given #21). 

This would just prevent more cases in which the user accidentally mutates the table and then gets confusing behavior on update.

src/transformer-components/transformerList.ts
```diff
@@ -627,6 +628,28 @@ const transformerList: TransformerList = {
       },
     },
   },
+  "Mutable Clone": {
```
thomascastleman, line 631: I wonder if this is the best name for this transformer--it's not clear to me why "clone" vs "copy" here, though I may just not be familiar with the established distinction between the two. That said, middle schoolers may not be either.

It might be nice if we chose something that conveys even stronger the fact that updates don't flow--I like "snapshot" for this reason but I understand it doesn't convey the mutability aspect very well. I'm not sure what would be best here. Thoughts? @stardust66 @P-bibs

thomascastleman (approved):
Looks good to me, just a few comments below.

P-bibs:
It appears if I apply a transformer to a dataset then the output is not editable as desired, but if I then apply `Editable Copy` to the output the copy is still not editable. Can you guys reproduce?

thomascastleman:
Oh, that actually makes sense since right now editable copy calls `uncheckedCopy` under the hood, which copies attributes exactly. So if the attributes have `editable: false` in the input (ie from the previous transformer), they will still be uneditable even after the editable copy is made. So it seems we want something like `makeDatasetImmutable` but that does the reverse and sets `editable` to true for all attributes, which we could apply to the result of the `uncheckedCopy`.

stardust66:
Yeah, I will fix that.

thomascastleman:
Nice, I think this looks good for now. If we can  get some better immutability features from the CODAP side, that would be great but I think we can merge this as what we can do for now.

P-bibs (approved):
Yep, looks good to me too.

PR #154, opened 2021-07-19
use function application syntax in dataset titles

thomascastleman:
This switches to dataset titles of the form:

- `Transformer(Dataset)` for transformers whose only parameter is a dataset (or 2)
- `Transformer(Dataset, ...)` for transformers which have other parameters (attributes, formulas, etc). The `...` indicates "more parameters went into this transformer, but you should check the description to read about them"
- something else for special cases.

Special cases include Partition, which creates multiple datasets and therefore seems to need something that differentiates its output datasets by their titles. Also, the single value transformers incorporate the attribute(s) they are computed over, because there is no "description" attached to text components.

Example:

Screenshot from 2021-07-19 11-45-24

### DIFF
--- a/src/transformer-components/util.ts
+++ b/src/transformer-components/util.ts
@@ -25,15 +25,13 @@ export async function applyNewDataSet(
 }
 
 /**
- * Returns the context's title, if any, or falls back to its name. Also
- * adds parentheses around the name if it determines the name
- * is not a single word.
+ * Returns the context's title, if any, or falls back to its name.
  *
  * @param context the data context to produce a readable name for
  * @returns readable name of the context
  */
 export function readableName(context: DataContext): string {
-  return parenthesizeName(context.title ? context.title : context.name);
+  return context.title ? context.title : context.name;
 }
 
 /**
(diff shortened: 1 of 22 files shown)


P-bibs:
How do we feel about removing the table title from the attribute created by running transforms? Since the title of the output table clearly communicates the source table, I don't think we need to duplicate that info in the attribute name. For instance, the below attribute name is a little obscene. If we also removed the table title from the tooltip it would help us implement the table name functionality we discussed this morning.

image

P-bibs:
I'm realizing the above comment doesn't have much to do with this PR and the discussion could be had elsewhere. I think the naming changes are great and the special-cases are handled well. I especially like that we managed to keep the partition information close to the beginning of the name.

P-bibs (approved):

thomascastleman:
I agree that is definitely obscene, and removing from the tooltip would make it easier to flow names, but recall that we chose to put it there in an attempt to communicate the "frozenness" of the running sum. That said, I'm not sure how well it does communicate that, and the length certainly feels like it detracts more than adds in clarity here.

stardust66 (approved):
Nice, looks good!

PR #155, opened 2021-07-19
allow multiple values in mode

thomascastleman:
This adds support for a single-value transformer producing a list of numbers, and applies it to the Mode transformer, which now produces all the mode values of the given attribute.


stardust66:
I'm wondering if it would be nicer to include square brackets and put spaces after commas in the output so that it's more obviously a list of values.

thomascastleman:
Here's what it looks like now: 

modes

stardust66 (approved):
Looks good!

PR #157, opened 2021-07-19
Error message tweaks

stardust66:
- Remove X emoji from CODAP eval error messages.
- Put quotes around context name in error messages.

### DIFF
--- a/src/utils/transformationDescription.ts
+++ b/src/utils/transformationDescription.ts
@@ -7,11 +7,13 @@ import {
   DDTransformerState,
   FullOverrideFunction,
 } from "../transformer-components/DataDrivenTransformer";
+import { readableName } from "../transformer-components/util";
 import {
   DataSetTransformationOutput,
   NumberTransformationOutput,
 } from "../transformers/types";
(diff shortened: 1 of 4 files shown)


src/utils/transformationDescription.ts
```diff
@@ -84,9 +84,9 @@ export function useActiveTransformations(
               .transformerFunction.kind === "datasetCreator"
           ) {
             setErrMsg(
-              `Error updating ${
+              `Error updating "${
                 (description as DatasetCreatorDescription).output
```
thomascastleman, line 88: This `description.output` seems to be the context name. Is it worth making a phone request here to `getDataContext` and then extracting the `readableName`? This would address one of Flannery's complaints.

thomascastleman (approved):
Looks good, just one comment below about names.

src/utils/transformationDescription.ts
stardust66: Yeah I think that's a good idea.

P-bibs (approved):

PR #158, opened 2021-07-20
several updates to the UI

thomascastleman:
- Make expression editor size smaller (it's about three lines vertically now)
- Make default height of _saved_ transformer windows smaller than the main transformer selector window (since they don't contain any of the save UI)
- Make the "Save This Transformer" UI toggleable and hidden by default
- Internally, renamed `TransformerSubmitButtons` to `TransformerSubmitButton` since there's one button now, and `TransformerSaveButton` to `TransformerSaveUI` since it's much more than just a button.

Notably, I was not able to get the "Apply transformer" and "Edit" buttons side-by-side on the saved transformer UI as Flannery requested, because the Apply button is actually inside a div that's part of one component and might have an error message in it also, and the Edit is part of another component. If you think you know a way to work around this, feel free to make the change. 

Here is what the hidden save UI looks like: 

hidden-save-ui


P-bibs:
Would it be possible to have the expression box start as one line and expand as the user writes more text? Flannery had made the comment that most formulas students write would probably be no more than a single line (sorry for not making that clearer in the minutes).

P-bibs:
How would we feel about not letting the user edit the name/purpose statement once a transformer is saved? I think it almost makes more sense this way, and it would allow us to put the `edit` button in the right place fairly easily.

thomascastleman:
I agree it would be ideal to have it expand. I can try to take a look at that.

As for editing name/purpose statements, I actually think it is important to keep them editable. If the purpose changes or you screw something up (or maybe you create the transformer without a purpose statement at first because it wasn't clear exactly what it was) it would be a pain if your only recourse was starting over.

P-bibs:
Fair point on the purpose statements/names. On a different note: should we bring in some icon library so we don't have to use `->` for type contracts?

thomascastleman:
Got the editor to fit the content height. I like this a lot better (it eliminates a lot of scrollbar space wastage).

The only other thing I can think of is filling the editor with a placeholder like "Write expression here..." but I don't know how to do this in Codemirror. Ideas @stardust66 @P-bibs ?

stardust66:
I think there's a codemirror addon for that. I can take a look at it?

stardust66:
Add this import
```javascript
import "codemirror/addon/display/placeholder.js";
```
Which allows you to use a `placeholder` field on the options object passed to CodeMirror. The resulting placeholder has the class `CodeMirror-placeholder` and we can style it separately.

thomascastleman:
Now, the expression editor has a placeholder that says "Formula expression" (lmk if you prefer something else).

Also, I added a close button from inside the popover since otherwise it can be inconvenient to close:

popover-with-close

P-bibs:
Small UI tweak:
image

thomascastleman:
Sleek, I like it. 

While we're on the subject of type contracts, how do we feel about `Row` being capitalized while the other types (following CODAP convention) are not? Should we maybe make it `row` to match?

P-bibs:
I had used `Row` to be in line with Bootstrap worksheets, but you're right that lowercase would be more consistent.

P-bibs:
Good to merge this?

thomascastleman:
Yup

PR #160, opened 2021-07-20
name changes for running aggregators

thomascastleman:
This removes the dataset name from running aggregator names, as was brought up in #154. It also gives Difference From a more useful tooltip.


thomascastleman:
To be clearer, it removes the dataset name from the _attribute names_ of running aggregators. The input dataset name is still present in the output dataset name.

PR #162, opened 2021-07-22
Show `[fixed]` labels on table with deleted inputs

P-bibs:
This PR adds to the context deletion listener to make sure that when a table is deleted all it's output tables get the string `[fixed]` appended to their title.

Notably, this title change does not flow through to further tables, so you can get the following scenario.
image
How do we think this case should be handled? Should we propagate the `[fixed]` notification to further tables, so we get `Copy(Copy(People)) [fixed]`, or should we propagate the new table name so we get `Copy(Copy(People) [fixed])`?

### DIFF
--- a/src/utils/transformationDescription.ts
+++ b/src/utils/transformationDescription.ts
@@ -18,7 +18,6 @@ import {
   updateText,
   notifyInteractiveFrameIsDirty,
   updateDataContext,
-  getDataContext,
   getComponent,
   updateComponent,
 } from "./codapPhone";
@@ -140,6 +139,8 @@ export function useActiveTransformations(
                 },
               });
             }
+          } else if (transformation.transformer === "Editable Copy") {
+            // If the transformer was editable copy then we don't have to do anything
           } else {
             if (transformation.outputType === TransformationOutputType.TEXT) {
               // If this is an SV transformer than update the output text component title
--- a/src/utils/transformationDescription.ts
+++ b/src/utils/transformationDescription.ts
@@ -17,6 +17,8 @@ import {
   notifyInteractiveFrameIsDirty,
   updateDataContext,
   getDataContext,
+  getComponent,
+  updateComponent,
 } from "./codapPhone";
 import {
   addInteractiveStateRequestListener,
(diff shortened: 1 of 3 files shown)


src/utils/transformationDescription.ts
```diff
@@ -104,6 +106,32 @@ export function useActiveTransformations(
...
+        }
+
+        // Rename transformations with newly missing inputs to add a [fixed] suffix
+        for (const transformation of transformationsWithNewlyMissingInputs) {
+          if ("output" in transformation) {
+            console.log("transformation", transformation);
+            const outputData = await getDataContext(transformation.output);
+            await updateDataContext(transformation.output, {
+              title: `${outputData.title} [fixed]`,
+            });
```
thomascastleman, line 131: It'd be nice to also update the description here (maybe just appending to the end) to give a more verbose explanation of what the "[fixed]" label means for this table.

src/utils/transformationDescription.ts
```diff
@@ -104,6 +106,32 @@ export function useActiveTransformations(
...
+          } else {
+            restOfTransformations.push(description);
+          }
+        }
+
+        // Rename transformations with newly missing inputs to add a [fixed] suffix
+        for (const transformation of transformationsWithNewlyMissingInputs) {
+          if ("output" in transformation) {
+            console.log("transformation", transformation);
+            const outputData = await getDataContext(transformation.output);
```
thomascastleman, line 128: I get an unhandled rejection for `getDataContext` when this is used with single-value transformers (I guess because they are not contexts but just text elements). Can you modify this to work with single-value transformers as well?

thomascastleman (commented):
Nice, this is neat. I'm not sure about propagating the label and perhaps we should ask sk. I suppose it depends on whether users consider it part of the table name or just as a piece of UI.

I think because Partition is an override it needs to be handled separately (right now it doesn't seem to add the fixed label to its output tables). I believe Editable Copy is an override as well so it may be useful to merge with main and handle both of those.

src/utils/transformationDescription.ts
P-bibs: Unfortunately, the API does not allow updating the description for some reason. I've reproduced this in the API tester: if you send an "update data context" request with a `metadata: {description: "test"}` field, then the request succeeds but the description isn't actually updated. I can look into this more but thankfully I don't think it's a deal-breaking issue.

thomascastleman: Ahhh yes I remember this--Jonathan confirmed that you can't currently update it through the API [here](https://codap.concord.org/forums/topic/accessing-dataset-description-through-plugin-api/). 

I guess this has implications for our ideas about flowing name changes since dataset names appear in descriptions and from that point on we just can't update them.

src/utils/transformationDescription.ts
P-bibs: Should be fixed now

thomascastleman:
Nice, Partition and aggregators look good now. 

Now what I'm noticing is that when we delete a context and update the titles, our dataset selector dropdowns are still displaying the old titles. This is probably because we don't receive a notification on title change, and `ContextSelector` is just using `useDataContexts`.

Could we manually cause an update, ie by calling `callAllContextListeners` or something?

P-bibs:
Ahh nice catch I hadn't noticed that. The `callAllContextListeners` approach definitely works. Is there a different way we should be doing it or is that good enough? I'll work on merging with main and adding support for editable clone.

thomascastleman:
Yeah not sure--Jason do you think there's a better way? @stardust66

P-bibs:
I think all the points that were brought up are addressed now
* Works with editable copy/partition
* Works with SV transformers
* Can't update description for now

Should we add something to the user documentation to explain `[fixed]` since we can't update descriptions?

thomascastleman:
Yeah that's a good idea--are there other corner cases or behavior that might be confusing that we should detail as well? We could have sections for those.

thomascastleman (approved):

stardust66:
> Yeah not sure--Jason do you think there's a better way? @stardust66

Since we just don't get notified, I don't think there's anything else we can do here.

stardust66 (approved):

PR #164, opened 2021-07-22
Check for missing value in running aggregators

stardust66:
Fixes the issue where running aggregators treat missing values as 0.

#### Before
image

#### After
image image

This is necessary because `Number` returns 0 for any whitespace string. I'm not sure using the previous value in the column is the right choice. Maybe it should be blank. We can discuss in the next meeting.

### DIFF
--- a/src/transformers/fold.ts
+++ b/src/transformers/fold.ts
@@ -66,11 +66,20 @@ function makeFoldWrapper(
   };
 }
 
+// Which value to put in the fold column if the current row contains a missing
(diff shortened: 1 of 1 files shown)


stardust66:
It occurred to me that we should probably be using `findTypeErrorsNumber` here.

thomascastleman:
This is definitely better behavior than treating it as 0, and we should probably incorporate into our broader discussion of missing values. 👍🏻

thomascastleman:
I still think we should bring these behaviors by sk just to check and be sure, but I think this is good to merge for now.

PR #165, opened 2021-07-23
Tests for flatten, groupby, and aggregators

stardust66:


src/transformers/__tests__/flatten.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  test("flatten of empty dataset with one collection", () => {
+    expect(
+      uncheckedFlatten({
+        collections: [{ name: "Cases", labels: {} }],
+        records: [],
+      })
+    );
+  });
```
P-bibs, line 44: Is there supposed to be a `.toequal()` here after the expect?

P-bibs (approved):

thomascastleman (approved):
Looks good

src/transformers/__tests__/flatten.test.ts
stardust66: Yup. Didn't realize.

thomascastleman:
Yeah--should we get this merged anyway though and come back to that when we've got the missing value infrastructure in place? It doesn't seem like we'll have missing values done before we let Bill and Jonathan start reviewing, so maybe we should just try to have all our tests so far merged and we can continue adding later.

stardust66:
Sure. Should I remove the tests that assume a certain behavior for missing values or just leave them in for now?

thomascastleman:
I think they're actually pretty close to what it seems like we're ending up with, so maybe leave them in. We can always take them out / remove later.

P-bibs:
You can always use `test.skip(...)` and leave a `TODO:` to add them back later

thomascastleman:
Can we merge into the main testing branch so it's all in one place?

stardust66:
I'm good with that.

PR #166, opened 2021-07-23
remove our language implementation

thomascastleman:
This removes `src/language` and any references to it. Sad to see it go but we're not using it at this point and as we transfer ownership of the code it would be confusing to leave it in.


P-bibs (approved):
Good call. Plus I'm a sucker for diffs with only deletions

stardust66 (approved):

PR #168, opened 2021-07-23
add support for recognizing dates as CODAP values

thomascastleman:
Now we get nicer errors with dates like this:

date


P-bibs (approved):
Wow I appreciate the regex work. We'll have to remember to change this (along with isBoundary, etc.) to use the CODAP type predicates once we have access them.

stardust66 (approved):

PR #170, opened 2021-07-23
Add importer button

stardust66:
Adds an "Importer" button that spawns the sheets importer.
image

On localhost CODAP the importer plugin doesn't work. I suspect it's because it's not hosted locally. Can we try and deploy this and see if this works? Then we can send the link.

### DIFF
--- a/src/transformer-components/TransformerREPLView.tsx
+++ b/src/transformer-components/TransformerREPLView.tsx
@@ -11,6 +11,7 @@ import { TransformerRenderer } from "./TransformerRenderer";
 import {
   getInteractiveFrame,
   notifyInteractiveFrameIsDirty,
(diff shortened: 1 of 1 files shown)


thomascastleman (approved):
This is just temporary right? For Rachel to work with? Because ultimately we don't want the importer to be married to transformers right

Looks good

stardust66:
Correct

thomascastleman:
We could also just have it launch an importer when transformers loads--that way there's no need for a button and she won't end up spawning multiple importers? But either works

stardust66:
True. I was thinking if she closed the importer she wouldn't have to reload to bring it back if there was a button.

P-bibs (approved):

PR #171, opened 2021-07-23
Check if selected context is a valid context

### LINKED ISSUE #169: deleted table used in transformer after deletion
To reproduce:

1. Transform the People dataset
2. Delete the People dataset
3. Hit "Apply Transformer" again without changing anything (the dropdown should be selecting one of the output tables of the previous transform)

Notice the transform applies to the People dataset, though it has been deleted and was not what was visible in the dropdown.

stardust66:
Fixes #169.

This is a similar approach to the one we take in the attribute selector.


P-bibs (approved):
:+1:

thomascastleman (approved):

PR #172, opened 2021-07-24
Add tests for pivot wider and pivot longer

P-bibs:


src/transformers/__tests__/pivot.test.ts
```diff
@@ -0,0 +1,161 @@
...
+  it("errors on non-displayable value in  valuesFrom attribute", () => {
+    expect(() =>
+      uncheckedPivotWider(TYPES_DATASET, "Boundary", "Number")
+    ).toThrow("Cannot use a boundary");
+  });
+  it("allows non-displayable value in  namesFrom attribute", () => {
+    expect(() =>
+      uncheckedPivotWider(TYPES_DATASET, "Number", "Boundary")
+    ).not.toThrow();
+  });
```
thomascastleman, line 57: These tests look good, but I think the attribute names (in the descriptions) are swapped--pivot wider errors on non-displayable values in `namesFrom` because they are used as attribute names, but allows them in `valuesFrom` since they stay values.

thomascastleman (approved):
Looks good

PR #174, opened 2021-07-26
Refactor project structure to ease migration

thomascastleman:
So far I've started by adding a `styles/` subdir in `ui-components/` and moving the `TransformerSaveUI` into `transformer-components/`.


P-bibs:
I've added some commits to update the project structure. Some caveats
* I haven't renamed anything yet. Just trying to focus on structure
* The `transformer-components` directory should be deleted, but that requires splitting apart the `types` and `utils` files and moving their pieces to the new relevant directories, which I don't want to do til we've settled on a structure.
* The imports are hideous, we can fix them later.

Other thoughts:
* I'm not sure what a good name for `transformer-template` is. My thought was it would contain `DataDrivenTransformer` and any helper components, so I suppose we should name it whatever we end up renaming `DataDrivenTransformer`
* One big question I have is where we should put `TransformerRenderer`. Right now it's in `transformer-template`, but that doesn't seem quite right. Thoughts?

thomascastleman:
- I really like having `views` with just two components which are the two views (REPL/saved)
- Putting `transformationDescription.ts` in `lib` also makes sense
- I agree it isn't clear where `TransformerRenderer` should go--but since it's effectively a wrapper around `DataDrivenTransformer` I think it's fine to have them in the same directory like that.

thomascastleman:
I am okay with this structure.

Naming-wise, here's what I think:

- `transformerList` --> not sure but maybe something like `transformerData`?
- `views/`
  - `SavedTransformerView` --> `SavedView`
  - `TransformerREPLView` --> `REPLView`
- `lib/`
  - `transformationDescription` --> something more like `transformerStore` but I'm not sure
- `transformer-template/`
  - `DataDrivenTransformer` --> `TransformerTemplate` (my argument for this being called a template is that it is not merely a component which configures the UI with props, it abstracts over the UI of _every_ transformer we have--so it feels much more template-y to me than others)
  - `TransformerSaveUI` --> `SaveUI` ? 
 
What do you think? @P-bibs @stardust66

P-bibs:
Looks good for the most part. The one thing that stand out to me is that I'm not sure `SaveUI` communicates enough information. I agree that since this is the `Transformers` plugin we probably don't need to include `Transformer` in component names for the most part, but even more than that if something is a component then we definitely don't need `UI` in it's title, which just leaves us with `Save`. What about `DefinitionCreator` or something of the like (reminiscent of the Pyret "definitions pane"). Then we could rename `SavedView` to `SavedDefinitionView` which I think is a better communication of what it is (`SavedView` is a bit generic).

thomascastleman:
That sounds good, I like the use of "definition" to be more specific. I can make those changes.

What do you guys think of `transformerList` and `transformationDescription` in particular? @stardust66 @P-bibs

stardust66:
I'm good with renaming `transformationDescription` to `transformerStore`, but I want to put `transformerStore` in another top level folder, maybe called `store`? Somehow I feel like it's not quite `lib` since it manages updating by itself.

thomascastleman:
I think moving it into a `store/` would make sense if we're going to split that one up into several files anyway.

thomascastleman:
Hmm but there are several components whereas only one "store" in terms of a store of transformer info and how they relate to each other.

P-bibs:
Ahh I was under the impression that this was a generic folder for this store and any we would need in the future that would then have a `transformerStore` specific folder. If it's only for this `store` could we do `store/`->`transformerStore/`  and then put the main functionality in `index.ts` along with helpers in `types` and `utils`?

thomascastleman:
Okay, I've split it up--feel free to look at that and reorganize if needed.

thomascastleman:
One thing worth asking is do we really think there'll be multiple stores? If not I think we should probably just have `store/` or `transformerStore/` at the top level containing the three `index.ts`, `types.ts`, and `util.ts` files, instead of nested directories.

thomascastleman:
OHHH lol is that what you meant the first time? Sorry if so, I totally misunderstood.

stardust66 (approved):
Looks good!

P-bibs (approved):
Haha yeah that's what I meant. All looks good now!

PR #176, opened 2021-07-26
Check whether output context has been deleted before adding [fixed]

### LINKED ISSUE #173: unhandled rejection after deleting transformed output
1. Transform an input dataset.
2. Delete the output dataset.

I get `Unhandled Rejection (Error): Failed to get context <output data context here>` from inside `getDataContext`.

stardust66:
Fixes #173.


thomascastleman (approved):
Nice, looks good.

P-bibs (approved):
Looks good to me. Could we add a comment on the `if`s to remind us why they're there so we don't accidentally reintroduce this bug?

PR #179, opened 2021-07-27
Improve transformer selector

P-bibs:
This PR adds a new transformer selector (using the `react-select` library) that allows searching for selectors by typing.

I know that this behavior exists with native dropdowns on most browsers, but I'm not sure most people know about, and I think the list of transformers is long enough that it becomes a pain to scroll through it each time.

Unfortunately, the new dropdown is quite a bit bigger, so I had to reorganize the "about" and "info" buttons accordingly.

image
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,7 @@ module.exports = {
   rules: {
     "dot-notation": "off",
     "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "^_" }],
     "no-constant-condition": ["error", { checkLoops: false }],
     "prefer-const": [
       "error",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,20215 @@
 {
   "name": "codap-flow",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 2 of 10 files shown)


stardust66:
Hmm, there might be some z-index issue with the placeholder text in the expression editor. To reproduce, pick "Build Column" and click on the selector again.

Screenshot from 2021-07-27 10-48-26

thomascastleman:
This is really nice!

I don't mind the select being larger than the others, since it's a special one. I do think, however, it might help to make it just a bit smaller than it currently is now (if possible).

Along that line, the Transformer Type header seems a bit too large as well right now (imo). I think it's fine if it's matching the larger dropdown, but still think it could be a bit smaller.

P-bibs:
@thomascastleman  I actually agree about the header being too big. I had to up the size because it didn't quite align with the "About" button in it's new location. If we keep the original size it looks like this:
image
That's actually not as better as I thought it would be, so I'm fine keeping the smaller size. I'll work on shrinking the dropdown.

P-bibs:
@stardust66 Nice catch. I'm noticing that if I disable the z-index style on the placeholder text then everything still appears to work fine. How would we feel about just adding a css rule along the following lines:
```css
.CodeMirror pre.CodeMirror-placeholder {
  z-index: unset !important;
}
```

stardust66 (approved):
Nice. Looks good to me!

thomascastleman (approved):

PR #182, opened 2021-07-27
Automatically select context if it is the only one

### LINKED ISSUE #136: auto-select a dataset if it is the only one in the document
From Jonathan:
> Quite a few, if not all, transformations have dataset selectors in the UI. I noticed that if you only have one dataset in your CODAP document you still have to make this selection. It would be helpful to the user to auto-select the one dataset in this case.

stardust66:
Closes #136.

This is implemented as part of `ContextSelector`.


thomascastleman (approved):
Nice, looks good.

P-bibs (approved):

PR #187, opened 2021-07-28
missing value reports

thomascastleman:
WIP for adding missing value reports to all transformers that may make suspect computations over missing values.

This extends the `DataSetTransformationOutput` and `SingleValueTransformationOutput` types to include a `MissingValueReport` of the following form:
```ts
type MissingValueReport = {
  missingValues: MissingValueLocation[];
  extraInfo?: string;
};

type MissingValueLocation = {
  collection: string;
  attribute: string;
  itemIndex: number;
};
```

The report is optional, though I think it might be useful implementation-wise to have transformers just always produce the report and have it ignored when it is empty (no `missingValues`).

I am prototyping this with just Mean at the moment and using `confirm()` to allow the user to either abort the computation when they notice it has missing values, or proceed with it in which a case the transformer output & report are generated.

I haven't tried adding this to updating because I think there are some questions there to address:

- Are MVRs updated when transformers update? Or is a new report generated? Or nothing at all?
- What if an MVR is deleted and then the input dataset is updated?

I'm also not entirely convinced of this way of indicating location--collection and attribute is good, but I feel like we need to explain the "item number" aspect and even then it will get confusing for hierarchical datasets.


thomascastleman:
Here's a screenshot of what it can look like now, using a table to present a summary of which values were missing. I had to include context info because some transformers consume multiple contexts and therefore would make it ambiguous.

missing-value-report

Also worth noting that although this looks nice, it is not the layout in which the components were thrown onto the CODAP workspace (which was quite spread out and random looking). This might have an effect on whether we keep this table view or revert to a single textbox with all the info.

thomascastleman:
Note that at the moment missing value reports do not interact with updating at all. I also want to demo the textbox + table behavior for sk to see what he thinks before merging.

thomascastleman:
Also worth noting: The missing value table for formula transformers does not contain collection/attribute info, but only indicates which rows of the input for which the formula evaluated to a missing value. This is because there is no one particular attribute/collection involved in the computation, since a formula can access any of the attributes across the table.

thomascastleman:
This good to merge? @stardust66 @P-bibs

stardust66 (approved):

P-bibs (approved):
Nice work on this! I appreciate how it fits into our existing dataset w/ description system for transformer functions

PR #190, opened 2021-07-28
Update type contract UI

P-bibs:
This PR adds functionality that lets the type selector take a single value instead of a list of options. This will display the value as a raw string instead of a dropdown. It also adds `Context:` before the dropdowns to give the user more instruction.

image
image


thomascastleman (approved):
Nice this is a lot better I think

PR #192, opened 2021-07-30
Use CODAP type checking

P-bibs:
Adjusts our formula type checkers to use the provided CODAP type predicates.

TODO: fix tests


src/lib/utils/typeChecking.ts
```diff
@@ -0,0 +1,77 @@
+import { CodapLanguageType } from "../../transformers/types";
+import { evalExpression } from "../codapPhone";
+import { prettyPrintCase } from "./prettyPrint";
+
+/**
+ * Enum representing all type checking predicates in CODAP
+ */
+enum CodapTypePredicateFunctions {
+  Boolean = "isBoolean",
+  String = "!isBoundary",
```
P-bibs, line 10: This is a bit suspect, but CODAP doesn't provide an `isString` so there's not much better we can do. Do you all have thoughts on other things that shouldn't be considered strings?

P-bibs:
We have two options when it comes to our transformer tests that check that type contracts are working properly:
* delete them, since we don't have access to the CODAP formula language in our tests and therefore don't have access to the type predicates.
* keep them and try to add type predicates to our `jsEvalExpression` function, knowing that these type checkers will have discrepancies with the CODAP ones and they will drift further out of line over time. However, this has the upside of letting us keep the tests.

Thoughts?

src/lib/utils/typeChecking.ts
thomascastleman: Hmm, yeah I guess this is tricky because string is sort of the catch-all category for CODAP. It still feels a little weird if we were to bake this in though as opposed to them providing it--maybe we should ask Concord folks about adding this.

thomascastleman:
Yeah, ultimately I would opt to remove them, since the entirety of the thing we're testing would be mocked by the mock, in which case we're just testing the mock, which doesn't seem all that useful.

thomascastleman (approved):

P-bibs:
Alright the relevant tests have been removed/updated. How does this look @thomascastleman  @stardust66

stardust66:
Do we want to ask the CODAP folks about `isString` or prioritize other things for now?

stardust66 (approved):

src/lib/utils/typeChecking.ts
```diff
@@ -0,0 +1,104 @@
...
+  values: unknown[],
+  evalFormula: (
+    expr: string,
+    records: Record<string, unknown>[]
+  ) => Promise<unknown[]>
+): Promise<number[]> {
+  const expr = `${predicate}(attr)`;
+  const records = values.map((value) => ({ attr: value }));
+  const results = await evalFormula(expr, records);
+  console.error(results);
```
thomascastleman, line 98: Should we remove this log?

thomascastleman (approved):
Yep looks good.

P-bibs:
I was thinking prioritize other things for now if that's all good

PR #193, opened 2021-07-30
Add loading indicator to apply transformer button

P-bibs:
I noticed that applying some transformers (especially if the table had boundaries) took long enough that it felt like nothing was happening for a moment when you hit `apply transformer`. I've added a loading spinner to help fix that. I opted for a loading spinner that appears on top of the button as opposed to the side to really reinforce the idea that the button is disabled while transformers are being applied. If you have alternate suggestions for stylng feel free to suggest them.
image

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,5 +1,5 @@
 
-.loading, .initError {
+.codap-connection-loading, .initError {
   font-size: 14px;
   margin: 15px;
 }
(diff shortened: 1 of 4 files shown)


stardust66 (approved):
Nice CSS animation. And good job on the centering!

thomascastleman (approved):
Nice, this is a small change that makes a big difference. I especially like how the cursor is disabled if you try to click the apply button while it's loading.

PR #194, opened 2021-08-02
removed [fixed] label

thomascastleman:
Deletes the functionality for adding the [fixed] label to outputs that have no way of updating.


P-bibs (approved):
Nice. I think this may fix #195 but we'll see

P-bibs:
Confirmed that this fixes #195

thomascastleman:
Ooh, it does seem to fix that

stardust66 (approved):
Yay. Looks good!

PR #196, opened 2021-08-02
Bring plugin to front when an error occurs

P-bibs:
Whenever `setErrMsg` is called with a non-empty error message, an api request is sent to select the plugin. This causes it to be brought to the front of the workspace so the user is made aware an error occured.


stardust66 (approved):
Nice. Looks good.

thomascastleman (approved):

PR #199, opened 2021-08-02
Autocomplete enhancements

stardust66:
- Complete attribute names that have spaces in them with backticks.
- When completing function names, move cursor between the parentheses after completion, just like the CODAP editor does.

### DIFF
--- a/src/components/ui-components/ExpressionEditor.tsx
+++ b/src/components/ui-components/ExpressionEditor.tsx
@@ -105,7 +105,20 @@ export default function ExpressionEditor({
       const completionList =
         word === ""
           ? []
-          : attributeNames.filter((w) => isLowerCasePrefix(w, wordLower));
+          : attributeNames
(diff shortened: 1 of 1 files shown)


thomascastleman:
For some reason I'm getting this error:

Screenshot from 2021-08-02 14-05-36

I made a multi word attribute and then tried to use it in the transform formula. It seems to think something is a function?

stardust66:
That's a CODAP side issue and it has been fixed. The error should go away if you update your local codap.

stardust66:
FYI, it's because it's compiling it to `multiplyFunction(c.vars.Multi Word Attribute, 10)` and not using square brackets.

thomascastleman (approved):
Awesome, it works now. These both really augment the usability of the expression editor. 👍🏻

P-bibs (approved):
:+1: Agree with Thomas, this makes the expression editor way better since no one's going to know they have to use backticks

PR #200, opened 2021-08-02
Shorten type error messages

P-bibs:
Rather than print the whole case that's erroring, we now only print the index. Any better suggestions? 

Old:
Screenshot from 2021-08-02 15-09-45
New:
Screenshot from 2021-08-02 15-08-10

### DIFF
--- a/src/lib/utils/typeChecking.ts
+++ b/src/lib/utils/typeChecking.ts
@@ -1,5 +1,5 @@
 import { CodapLanguageType } from "../../transformers/types";
-import { prettyPrintCase } from "./prettyPrint";
+// import { prettyPrintCase } from "./prettyPrint";
 
 /**
  * Enum representing all type checking predicates in CODAP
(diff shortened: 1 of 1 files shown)


thomascastleman (commented):
Hmm, the former feels more informative to me, but if shorter is a must then the latter does the job.

Only comment is, should we 1-index it? I have been referring to cases with 1-indexing elsewhere

stardust66:
We need to put the new message somewhere in the error messages doc. https://docs.google.com/document/d/1YsAyYy6L7hC2EPqVWpgM6bwzJiaibu6Jtso1RAV7tf8/edit#

P-bibs:
I'm actually not sure if we need to make them shorter. That was just a note that was added during our meeting with Joe at some point. If we think they're fine as is then we can make this  a no op.

P-bibs:
Closing this since #206 removes the need to have shorter error messages

PR #201, opened 2021-08-02
Extract error message strings

stardust66:
- Use `i18next`, an internationalization library.
- Extract all error messages from components into a single file.

This will also make translation in the future easier. In the short term it helps with keeping track of all our error messages.


thomascastleman (approved):
Nice, I like how this removes a lot of redundancy among the transformers with similar UI checks. 

Have the changes from the Google doc already been applied or are we waiting on more feedback there?

P-bibs (approved):
Looks good to me! (or should I say sieht gut für mich aus?)

P-bibs:
Let me know if any of the merge conflicts take too long and you need a second hand

stardust66:
> Have the changes from the Google doc already been applied or are we waiting on more feedback there?

I only applied one change. I'm not sure if Emmanuel's done with reviewing yet.

Issue #186, opened 2021-07-28
text components not properly removed from updating once deleted


thomascastleman:
To reproduce:

1. Do a single-value transformation on a dataset.
2. Delete the output textbox.
3. Change the input table.

I get an unhandled rejection: `Failed to get context <text context name>`

thomascastleman:
This appears to be the call to `getDataContext` that is erroring:
https://github.com/P-bibs/codap-transformers/blob/6d62438a825731d1e7a0c3ad6096c43b663e0e2a/src/transformerStore/index.ts#L92-L94

This error handling currently assumes the transformer produced a dataset output (since it tries to get the context) and so is erroring on text producing transformers.

thomascastleman:
The underlying issue however seems to be that text-producing transformations aren't removed when the output is deleted. It seems like we would need something along the lines of 
https://github.com/P-bibs/codap-transformers/blob/2b945ecf27b074f40790d9f68edf2042257a223a/src/lib/codapPhone/listeners.ts#L209-L213
but for text components.

thomascastleman:
Note that the call to `getDataContext` causes issues anytime an error occurs during a text update

PR #202, opened 2021-08-02
fixed updating errors for text outputs causing crash

thomascastleman:
This addresses part of #186 -- the unhandled rejection part where we tried to get a text component as a data context.

I also made it so that new updates cause the error message to clear before the update happens. That way, if an update error is fixed, the flowing through of the update will remove the error message.

### DIFF
--- a/src/components/transformer-template/TransformerTemplate.tsx
+++ b/src/components/transformer-template/TransformerTemplate.tsx
@@ -329,7 +329,7 @@ const TransformerTemplate = (props: TransformerTemplateProps): ReactElement => {
           type: ActiveTransformationActionTypes.ADD,
           newTransformation: {
             inputs,
-            extraDependencies: [],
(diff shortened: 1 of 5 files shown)


stardust66 (approved):
Thanks for fixing this!

P-bibs (approved):

thomascastleman:
This now listens for text deletions and removes the corresponding transformations from the store.

It relies on [this patch](https://github.com/stardust66/codap/commit/964ca39c60bfb68aa28e30bd896b08f34e75734e) to the CODAP repo which adds the name to the deletion notification for all components.

It should fully fix #186 though now I'm worrying about interactions between undo/redo and updating of text components. That's for a separate PR I think.

Look good to merge now that this has the full fix? @stardust66 @P-bibs

P-bibs:
Yep go for it :+1:

PR #204, opened 2021-08-03
invalidate cache for deleted contexts

### LINKED ISSUE #132: Deleted datasets are still cached

thomascastleman:
Fixes #132, I think. I can't reproduce the erroring behavior mentioned there and I think it may be because our UI selectors have since gotten smarter about dealing with deleted datasets.

@stardust66, does this do everything necessary to invalidate the cache for the deleted context? Is there anything else I'm missing?


stardust66 (approved):
Yup. This looks good!

PR #205, opened 2021-08-03
Make default join an inner join and add outer join

stardust66:
Significantly changes the behavior of join.

1. The default join becomes an inner join, where the only cases that are kept are the ones where the joining attributes match.
2. Adds outer join, of which there are two types. The left (outer) join keeps all the values from the base dataset, and only the cases from the joining dataset that match. The full (outer) join keeps all values from both datasets.
3. This also changes the way we deal with duplicate matches. We no longer only consider the first match, but include all matches by duplicating cases. This is in response to Joe's feedback. See in the recording at around 27:25. Basically, if a person has multiple grades, we would probably want to include all of them in the result and not just take the first one.

#### Example of an Inner Join
image


thomascastleman (approved):
Wow, I am impressed. The original join was supposed to mirror CODAP behavior, but I like these joins a lot better. Nice work!

thomascastleman:
@stardust66 Can I ask you to update [the docs](https://docs.google.com/document/d/1NZA9gxtu6jD3M-5SQyx0tvV2N5qYKMgRm1XUwMnLgJU/edit) by deleting the previous Join and adding the descriptions of the new inner/outer joins (with example screenshots)?

Don't worry about linking to it from the inline docs yet (unless you want to), I think things will be shifted around anyway and not sure if the links will remain valid.

P-bibs (approved):
There's even tests :heart_eyes:. I'm with Thomas this looks great

PR #206, opened 2021-08-03
Add initial error store implementation

P-bibs:
This PR adds an alternate error tracking mechanism that allows multiple errors to be displayed at once.


P-bibs:
Should be good to go now. If you want to test it out you can do the following:
1. GroupBy an attribute on the microdata portal
2. Switch to another transformer and try and hit "apply transformer" without filling out any info to get an error.
3. Delete the attribute you GroupedBy in the microdata portal.
You should now have one error from step 2 and a separate error form step 3.

One thing that doesn't work quite right: generating two tables with the same transformer without switching transformers ties them both to the same errorId. Therefore, they will squash each other's errors. What this means in practice is that if you make two table with the same transformer right in a row then you'll only see one of their errors at any given time.

stardust66:
Good job! I like this system much better. It actually works for multiple errors.

I have a suggestion about error IDs. Instead of switching IDs when the transformer type changes, why not have a fixed ID for when the user is working on the transformers pane and give each new transformation, when it goes through without errors, a new ID? This way, errors won't be squashed.

thomascastleman (commented):
Nice, this seems much better than letting all new errors cover up others (even given the behavior you describe).

Some thoughts:
- Can the `ERROR_MESSAGE_CUTOFF_LENGTH` be bigger? I find that some pretty short messages are being cut off. I think if we can experiment and get it to display ~2 lines of text (given the default width of the plugin) before cutting off that would be ideal.
- On the UI end, I love the ability to switch back and forth and close the message, but it feels weird that the X button is at the bottom. It probably is good that the X and the arrows are at the same level, so could we try flipping it and having those occur at the top with the message below? This would also separate it from the "read more" button which I think might help.
- Some error messages are quite generic, and the fact that they now persist can lead to confusion about which transformer they are tied to. For example, try going through transformers and just hitting apply--you'll get lots of "Please choose a valid dataset to transform", but they don't indicate which transformers gave the errors. I know this is a contrived example but I feel like a similar situation might happen in practice. Is there a good way we could link the messages to the transformers that raised them?
- Clearing: it seems like certain kinds of errors should just clear when a different transformer is selected (in fact, maybe most, except updating), since the inputs clear. I.e., if I didn't choose an attribute, then I switch to another transformer, it seems weird that the attribute error persists. Maybe clearing more (except update errors) could help with the above bullet point as well

P-bibs:
@stardust66 good call. I implemented this and things appearing to be working as expected now, with each transformer being able to display it's own errors.

@thomascastleman Good catch on the error length. I originally had it at 100 but set it to 50 for debugging and forgot to change it back. I'll work on your other suggestions and see what I can come up with.

P-bibs:
Here are a few different UI options. Thoughts on which is best?

Screenshot from 2021-08-05 10-30-24

Screenshot from 2021-08-05 10-44-08

Screenshot from 2021-08-05 10-46-07

thomascastleman:
My vote is for 1.

P-bibs:
@thomascastleman I added a hook that clears REPL-related errors when switching transformers. Does this address your last two bullet points?

thomascastleman:
Yep! That looks good.

One minor thing is that when you cause an update error, then make a change that fixes the error (and the update goes through), the update errors sticks around. Ideally we could clear updating errors when the update goes through successfully again, but I don't know how feasible this is...

I definitely think this is an improvement over the previous system

thomascastleman (approved):

P-bibs:
Can you guys take a close look at 5b4920a . It's just one line that should fix what Thomas is saying about update errors not getting cleared when updates go through successfully, but it's a pretty important part of the code and I want to make sure there aren't any unintended side effects I'm not noticing.

thomascastleman (commented):
Hmm this seems to work but I don't understand why--why aren't we clearing the error before trying to do the update? It seems like this would squash any errors that happen in the update (it doesn't but I don't see why).

P-bibs:
I'm pretty sure it's because if an error is raised then the `try` block gets bailed out of and the line that clears the error is never reached. Regardless, I agree that the logic would make more sense if we cleared the error before the update.

thomascastleman:
Ohhhh right, it makes sense now. Yep this looks good!

stardust66 (approved):
Nice. This looks good to me, and the updating seems to work.

PR #208, opened 2021-08-04
add a prompt to the filter expression

thomascastleman:
Let me know thoughts on the specific wording of the prompt. Right now it is "Keep all rows that satisfy:"

We can also use this prompt feature for other expression editor transformers. I think Sort might benefit from it, but I'm not sure the right wording to use.

### DIFF
--- a/src/components/transformer-template/TransformerTemplate.tsx
+++ b/src/components/transformer-template/TransformerTemplate.tsx
@@ -68,7 +68,7 @@ interface DropdownInit extends ComponentInit {
     value: string;
   }[];
 }
-interface ExpressionPromptInit extends ComponentInit {
+interface PromptInit extends ComponentInit {
   prompt: string;
 }
 interface ExpressionInit extends ComponentInit {}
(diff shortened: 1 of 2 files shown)


thomascastleman:
Here's what the prompt looks like btw:

filter-prompt

src/components/transformer-template/TransformerTemplate.tsx
```diff
@@ -88,6 +91,8 @@ export type TransformerTemplateInit = {
   textInput2?: TextInputInit;
   dropdown1?: DropdownInit;
   dropdown2?: DropdownInit;
+  expressionPrompt1?: ExpressionPromptInit;
+  expressionPrompt2?: ExpressionPromptInit;
```
P-bibs, line 95: Since the way these are implemented they're just generic text that's inserted into the component and don't have to specifically be next to an expression, what if we just called them `prompt1` and `prompt2` or `text1` and `text2`?

P-bibs (approved):
I think the text is great. Just one comment about implementation. Not sure about sort, maybe "values to sort by" or "keys to sort by" or "formula to generate keys to sort by"? They all seem too clunky

stardust66 (approved):
Nice, this looks good.

PR #209, opened 2021-08-04
use icon for about button

thomascastleman:
This tries out using an `IconButton` for the "About" button:
| Button      | IconButton |
| ----------- | ----------- |
| button | about |

I do like that this reduces the size of the button but it has the disadvantage of being quite similar to the (i) icon and right above it. Which do you all prefer?

### DIFF
--- a/src/components/info-components/AboutInfo.tsx
+++ b/src/components/info-components/AboutInfo.tsx
@@ -8,7 +8,15 @@ function AboutInfo(): ReactElement {
   return (
     <Popover
       button={
-        <IconButton style={{ padding: "0" }} size="medium">
+        <IconButton
+          style={{
+            padding: "0",
+            position: "fixed",
+            bottom: "15px",
+            right: "15px",
+          }}
+          size="medium"
+        >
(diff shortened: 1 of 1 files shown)


P-bibs:
I wonder if there's a better place we could put it so user's don't get confused and they can tell which button to press if they want info about a specific _transformer_. What if we put the about icon in the bottom right of the plugin?

stardust66:
I think putting it at the bottom right is worth a try. I agree that having (?) above (i) is not ideal.

thomascastleman:
Bottom right:
bottom-right-1
With Save UI open (and scrolled):
bottom-right-2

Thoughts? @stardust66 @P-bibs

thomascastleman:
Using `position: fixed` so it scrolls with you and doesn't go out of sight.

stardust66:
I like this better.

P-bibs:
I think I agree. Can you push changes so I can test it out?

P-bibs:
Yep looks good

PR #214, opened 2021-08-05
Invalid cache on attribute rename

stardust66:
This fixes a bug where if you do a transform on an attribute and that attribute is renamed, the update will not error because it still sees the old name for the attribute in the cache.


thomascastleman (approved):
Good catch

P-bibs (approved):

PR #217, opened 2021-08-05
make partition updates synchronous

### LINKED ISSUE #95: updating many contexts causes crash
- Create several (>6 or so) transformed datasets off of one input dataset
- Apply an update to the input table

I get:
Unhandled Rejection (TypeError): responses is not iterable

This happens in `updateContextWithDataSet`, it looks like `responses` is `undefined` in 
```ts
const responses = await callMultiple(requests);
```

@stardust66 Any idea what's happening here?

thomascastleman:
This fixes #95, at least for partition. However, I have no idea why. It does so by making partition updates synchronous. As a consequence this makes partition updates a lot slower.

I also threw in a more specific error if the response from CODAP comes back `undefined` in `updateContextWithDataSet`, so that we'll at least get something better than responses is not iterable.

I'm not sure what we want to do about this since I'm the only one who can reproduce.

### DIFF
--- a/src/lib/codapPhone/index.ts
+++ b/src/lib/codapPhone/index.ts
@@ -767,12 +767,6 @@ export async function updateContextWithDataSet(
 
   const responses = await callMultiple(requests);
 
-  if (responses === undefined) {
-    throw new Error(
-      `Received no response from CODAP when updating ${contextName}`
-    );
-  }
-
   for (const response of responses) {
     if (!response.success) {
       throw new Error(`Failed to update ${contextName}`);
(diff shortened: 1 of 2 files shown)


stardust66:
I'm also not sure what to do about this, but it doesn't seem like a big change anyway. I also don't see a big slow-down, so maybe we can just merge the fix.

stardust66 (approved):

P-bibs (approved):
Hmm I poked around a bit but also can't figure out why this would work. Anyway, I agree with Jason the performance isn't noticeable so looks good!

thomascastleman:
I think per Jason's explanation of iframe-phone returning `undefined` sometimes when serving a request takes too long, it may be because asynchronous partition updates happen too fast. Making them synchronous seems to slow it down enough that iframe-phone doesn't return `undefined`.

PR #218, opened 2021-08-05
check all responses from iframe-phone

thomascastleman:
This adds checks to every `response` in a call to iframe-phone to ensure it is not `undefined` before using it. In the `undefined` case it gives a more informative error message about an invalid response than what would be raised if we just used the value.

### DIFF
--- a/src/lib/codapPhone/index.ts
+++ b/src/lib/codapPhone/index.ts
@@ -114,12 +114,16 @@ export async function updateInteractiveFrame(
         values,
       },
       (response) => {
-        // NOTE: Ensure the response exists, since if this is run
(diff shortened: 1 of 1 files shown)


stardust66 (approved):

P-bibs (approved):
Looks good! Do we need to put the error in the strings file?

thomascastleman:
I didn't but only because the other errors in `codapPhone/` aren't--should we migrate them all over? @stardust66 Or was there a reason you left those

stardust66:
Yeah at first I didn't include those because I thought that codapPhone could eventually be decoupled from transformers, but honestly it's not happening soon so it makes more sense for the errors to go in the strings file now.

PR #219, opened 2021-08-05
Flow title updates

stardust66:
Add `useEditedOutputs`, which keeps track of the contexts that have had their titles manually edited, and use this information to decide whether to include updated titles in normal transformer updates.

### DIFF
--- a/src/components/transformer-template/TransformerTemplate.tsx
+++ b/src/components/transformer-template/TransformerTemplate.tsx
@@ -199,7 +199,7 @@ const titleFromComponent = (
   return tmp && tmp.title ? <h3>{tmp.title}</h3> : <></>;
 };
 
-interface DatasetCreatorFunction {
+export interface DatasetCreatorFunction {
   kind: "datasetCreator";
   func: (state: TransformerTemplateState) => Promise<TransformationOutput>;
 }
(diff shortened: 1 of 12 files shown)


stardust66:
So, it turns out that what was causing the issue was not that we can't tell apart manual edits and plug-in updates to the titles, but that we get a title edit notification when we create a new context, which messed things up from the start. This issue is solved for all transformations by checking whether the updated title belongs to the dependencies of any active transformation whenever we get a title update notification, so we only keep track of title edits for contexts after we've created them through a transformer.

thomascastleman:
Awesome! This will eliminate a lot of confusion I think and allow people to give their datasets meaningful names.

My only thought for possible improvement has to do with the dataset description--previously we couldn't update it for some CODAP reason but I asked Jonathan and I think he's fixed this in the latest build (see [this report](https://www.pivotaltracker.com/n/projects/1055240/stories/179027655)).

Would it be possible to incorporate description updating? This would be great, just so the title/description remain in sync and we don't get the following situation which might cause confusion:

description

thomascastleman (approved):
Nice! Seems to work great

P-bibs (approved):

PR #220, opened 2021-08-06
Put codapPhone strings in the strings file

stardust66:


thomascastleman (approved):
Looks good. Should we also move this error? 
https://github.com/P-bibs/codap-transformers/blob/234521623cb6ff2785a8ecd79284a170813d1eba/src/lib/codapPhone/error.ts#L6

stardust66:
Oh right. Good catch!

PR #221, opened 2021-08-06
Disable CODAP close button and add custom close button

P-bibs:
This PR disables the CODAP close button for the plugin component and adds a custom close button that asks the user to confirm if they close the plugin while there are any `activeTransformations`. This also applies to saved transformers

### DIFF
--- a/src/views/util.ts
+++ b/src/views/util.ts
@@ -5,7 +5,8 @@ export async function closePlugin(
   activeTransformations: Record<string, TransformationDescription[]>
 ): Promise<void> {
   if (
-    Object.keys(activeTransformations).length === 0 ||
+    // If there aren't any active transformers then we don't need to confirm
+    Object.values(activeTransformations).some((a) => a.length > 0) ||
     confirm(
       `Closing the transformers plugin will stop transformed datasets from updating. Are you sure you'd like to proceed?`
     )
--- a/src/views/util.ts
+++ b/src/views/util.ts
@@ -6,7 +6,8 @@ export async function closePlugin(
 ): Promise<void> {
   if (
     // If there aren't any active transformers then we don't need to confirm
-    Object.values(activeTransformations).some((a) => a.length > 0) ||
+    Object.keys(activeTransformations).length === 0 ||
+    !Object.values(activeTransformations).some((a) => a.length > 0) ||
     confirm(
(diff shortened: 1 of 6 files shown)


src/views/util.ts
```diff
@@ -0,0 +1,16 @@
+import { deleteComponent, getInteractiveFrame } from "../lib/codapPhone";
+import { TransformationDescription } from "../transformerStore/types";
+
+export async function closePlugin(
+  activeTransformations: Record<string, TransformationDescription[]>
+): Promise<void> {
+  if (
+    Object.keys(activeTransformations).length === 0 ||
```
stardust66, line 8: This doesn't quite work to check if there are any active transformations, since the keys are just input names, and are not deleted when transformations get deleted. I suggest

```typescript
function hasActiveTransformation(activeTransformations: ActiveTransformations) {
  return Object.values(activeTransformations).some((a) => a.length > 0);
}
```

Everything else looks good!

thomascastleman:
Nice, things look good on the functionality front.

UI-wise, I have two thoughts:

1. The "Close Plugin" button feels a bit clonky with how large it is. Is there a good reason not to use an icon (and have a tooltip that communicates "close plugin")?
2. Related to 1., having the words "Close Plugin" on a saved transformer feels a bit weird to me from the student perspective--we know that they're indeed plugin instances, but to the user, these are just saved transformers. I don't think that phrase will make much sense to them. Changing to an icon would also fix this (though we might want the tooltip to switch between "Close plugin" and "Close definition" or something). 

If we decide to go with an icon, it might be an opportunity to try putting the about icon back up at the top, maybe to the left of the close icon? That way there's sort of a menu of icons with tooltips.

P-bibs:
Yep I agree about the button being awkward UI-wise. I'll try some of those fixes and see how it looks.

src/views/util.ts
P-bibs: Good catch!

P-bibs:
How do we feel about this? I think the "L" shape that the icon buttons make is a little strange but not dealbreaking
Screenshot from 2021-08-06 16-02-22

thomascastleman:
I do like this better. I think it's a better use of screen real estate.

Looks good to me!

thomascastleman (approved):

stardust66 (approved):
Nice. Looks good!

PR #228, opened 2021-12-23
Capitalize type names in contracts

### LINKED ISSUE #223: Capitalize type names in contracts
A convention we inherited from HtDP and which lives on in Bootstrap is that type names are always capitalized.

image

thomascastleman:
Fixes #223. 

Worth noting that one reason we may have stuck with lowercase types is that it matches the uses of type names in the CODAP language (see functions `number()` or `boolean()` for instance).

### DIFF
--- a/src/transformerList.ts
+++ b/src/transformerList.ts
@@ -34,6 +34,7 @@ import { partitionOverride, partitionUpdate } from "./transformers/partition";
 import { editableCopyOverride } from "./transformers/editableCopy";
 import { transformAttribute } from "./transformers/transformAttribute";
 import { compare } from "./transformers/compare";
+import { codapLanguageTypes } from "./transformers/types";
 
 export type TransformersInteractiveState = {
   transformerREPL?: {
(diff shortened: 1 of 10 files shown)


src/transformerList.ts
```diff
@@ -124,6 +124,9 @@ function docLinkFromHeadingID(headingID: string): string {
   return `https://docs.google.com/document/d/1NZA9gxtu6jD3M-5SQyx0tvV2N5qYKMgRm1XUwMnLgJU/edit#heading=${headingID}`;
 }
 
+// All possible types, used in user-configurable contracts.
+const allTypes = ["Any", "String", "Number", "Boolean", "Boundary"];
```
thomascastleman, line 128: There's a way of extracting this list from the `CodapLanguageType`, right? @stardust66 @P-bibs 

If we can, we could have that be the source of truth for what types exist and not duplicate it here.

P-bibs: I pushed a possible fix. You have to do it the other way around, declaring the array and the deducing the type from that. Further, the array has to be constant, which means I had to sprinkle some `readonly`s around. If we decide we don't like the fix I can revert it.

thomascastleman: Looks good to me! I don't have a super strong preference between the two.

Issue #230, opened 2021-12-30
confusion about type contracts


thomascastleman:
Emmanuel indicated there was some confusion over what type contracts apply to, a formula or the transformer itself. Currently, contracts only appear over formulas and apply to the formulas and not transformers.

Is there a way we can make this fact clearer through the UI? Filter I think does the best job of this:

![filter-ui](https://user-images.githubusercontent.com/13399527/147777644-261a8189-54aa-4be4-9fb1-1728c3dd6513.png)

The leading phrase "Keep all rows that satisfy" makes it clear what the expression is for. Should we add a similar phrase for other transformers that use expressions?

thomascastleman:
@schanzer @shriram to weigh in

shriram:
Oooh, that would be really nice if it could be done! It would also make each transformer more self-descriptive, reducing the need to ever look at documentation (which is always a win).

schanzer:
@thomascastleman I think this is an excellent idea. Here's what I would put for the transformers:

**Transform**
For each row, replace the value of attribute `xyz` with the result of the expression:

**Build**
For each row, add the attribute `xyz` with the result of the expression:

My head isn't wrapped around the others as much, since we don't use them at all (yet!). Curious what you'd all propose for those?

thomascastleman:
The only others that use formulas are Sort and Reduce:

- Sort has a single formula which is evaluated at each row and that value is used to sort the rows.

- Reduce has two formulas, one which determines the starting value of the accumulator, and another which is the formula for computing the next value of the accumulator given the current value.

schanzer:
`Sort` - Sort the rows in this table by a value, which is computed by: <blank>
`Reduce` - (1) Starting with: <blank>, (2) compute the next value for each row by: <blank>

thomascastleman:
Closed by #235

PR #234, opened 2022-01-10
Formula prompts

thomascastleman:
Addresses #230 by adding prompt phrases for all formulas (similar to Filter's "Keep all rows that satisfy...").

Here's an example of what this looks like now:
prompts

The prompt is the phrase in italics preceding the formula. Hopefully this will make the role of formulas extra clear.

Note: Emmanuel requested prompts like "For each row, replace the value of attribute ____ with the result of the expression:" (Transform Attribute) where the _____ is filled in with the attribute being operated on. I've implemented a simpler version with static prompt strings which don't contain info that depends on the input, but we might be able to extend this to do so.


stardust66 (approved):
This looks good to me!

P-bibs:
Look good :+1:. I tried out a more dynamic extension on #235 if you want to take a look. It's a bit hacky so I didn't fully implement it, but if we're satisfied with the concept then I can follow through.

thomascastleman:
Closing in favor of #235

PR #235, opened 2022-01-10
Add dynamic-ness to formula prompts

P-bibs:


thomascastleman:
Nice, this looks great! I'd be happy to go with this version in place of #234.

PR #237, opened 2022-01-26
Name purpose UI

### LINKED ISSUE #231: name + purpose statement UI
Some of the feedback we got was to match the design recipe template as close as possible. For Bootstrap purposes, would it help if the dropdown labeled "Save This Transformer" were moved up to the top of the UI, just below the transformer dropdown?  This could make name / purpose statement more explicitly the initial steps before filling out the "body" of the transformer.

@schanzer @shriram

stardust66:
Resolves #231.

Transformer with formula:
image

Transformer without formula:
image

Things to tweak:
- Should purpose statement have its own heading or be included in the expression input?
- We could remove the toggle for the save button since it's just a single button now.

Edit: Updated screenshots.

### DIFF
--- a/src/components/transformer-template/DefinitionCreator.tsx
+++ b/src/components/transformer-template/DefinitionCreator.tsx
@@ -1,17 +1,7 @@
-import React, { ReactElement, useEffect, useState } from "react";
-import { TextArea, TextInput } from "../ui-components";
+import React, { ReactElement, useState } from "react";
 import { BaseTransformerName } from "../../transformerList";
 import { SavedTransformerContent, TransformerSaveData } from "./types";
-import {
-  createDataInteractive,
-  getInteractiveFrame,
-  notifyInteractiveFrameIsDirty,
-} from "../../lib/codapPhone";
-import {
-  addInteractiveStateRequestListener,
-  removeInteractiveStateRequestListener,
-} from "../../lib/codapPhone/listeners";
-import { InteractiveState } from "../../lib/codapPhone/types";
+import { createDataInteractive } from "../../lib/codapPhone";
 import "./styles/DefinitionCreator.css";
 import ErrorDisplay from "../ui-components/Error";
 import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
@@ -29,16 +19,12 @@ export default function DefinitionCreator({
   base,
   disabled,
 }: DefinitionCreatorProps): ReactElement {
-  const [currentName, setCurrentName] = useState<string>("");
-  const [description, setDescription] = useState<string>("");
   const [saveErr, setSaveErr] = useState<string | null>(null);
   const [saveUIShown, setSaveUIShown] = useState<boolean>(false);
 
-  function saveTransformer(
-    name: string,
-    description: string | undefined,
-    data: TransformerSaveData
-  ) {
+  function saveTransformer(data: TransformerSaveData) {
+    const { name, description } = data;
+
     if (name.trim() === "") {
       setSaveErr("Please give the transformer a name before saving.");
       return;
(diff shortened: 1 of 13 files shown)


retabak:
Hello! My first impression is that this looks great - but I'm wondering if there's a link that I can follow to try interracting with this new UI a bit? Let me know.

stardust66:
@retabak Hi! Just deployed it here: https://codap.concord.org/app/static/dg/en/cert/index.html?di=https://statuesque-souffle-9d7a34.netlify.app/

thomascastleman:
Looks good to me, just a couple things:
- I think we should get rid of the dropdown for the save button, as now it's the only thing there. We can probably just make it a "Save Transformer" button at the bottom that's always visible
- The "Purpose Statement" header above the purpose statement textbox feels redundant, given the placeholder text says the same thing. Can we just remove it?

thomascastleman:
The other thing that might need to change is that the location of the purpose statement hasn't moved for saved transformers - I think we probably want those to agree so folks don't get confused if they've written a purpose statement for a formula and then it appears near the top

thomascastleman:
The latest changes look great - I think the last thing to look at is the redundancy of placeholder text in some of the text boxes. Given Rachel's feedback, do you think we could do something like #242 but for the Transformer name input? Then we could suggest names that would be relevant to the Transformer being saved.

For purpose statements, I think we can just have a stock placeholder, but need to be careful to adapt it depending on if it's a purpose statement for the Transformer or for the formula - for a Transformer, maybe "What does the Transformer do to its inputs?" and for a formula, "What does the formula do for each row?" as Rachel suggests.

stardust66:
For this I'm considering making the purpose statement configurable through `transformerList`. Having to dynamically place the purpose statement seems brittle. Do you think this is worth doing?

thomascastleman:
I agree making it part of `transformerList` makes sense.

I also agree about trying to eliminate the scroll - we could also maybe get closer to this by reducing the vertical size of the purpose statement box and putting the Apply and Save buttons side by side.

thomascastleman:
I think this looks good to merge, the only last thing I think we should change is the Transformer name placeholder text. I think we should make it part of the `transformerList` configuration and have it be custom to each Transformer.

@retabak Could you give us a list of what you'd like to see for placeholder text for the name of each Transformer? It'll look something like this:
Screenshot from 2022-06-02 12-58-03

P-bibs:
I noticed one change from the previous functionality that may have been inadvertent: when saving a transformer, we no longer clear the `name` and `purpose statement` fields. Was this on purpose or should we restore the original functionality?

stardust66:
Good question. I think since the name and purpose statement fields are more integrated in the body of the transformer now, and we don't clear any other field when we save, we should either clear every field on save or leave them alone.

thomascastleman:
Hm, I could imagine a case being made for clearing the name still because we don't allow Transformers to have duplicate names, so if you were to hit save again it would immediately fail. That said, I don't think it's necessarily better to have it fail because we've cleared the name/purpose statement, and I could imagine a use case of saving several similar Transformers, in which a case having both of those fields persist and being able to just tweak them might be desirable.

So my vote is to leave them alone.

thomascastleman:
I just pushed placeholder text for the transformer name input - I added it to the `transformerList` but it's a little awkward because the `title` of the input is always the same (just "Transformer Name") - it's the `placeholder` that's changing. I didn't know a way to give the title a default value or otherwise not have to specify it for each transformer, so let me know if you have a workaround for that @stardust66 @P-bibs

stardust66:
What I did for the purpose statement was to not have it inherit from `ComponentInit`. We can do the same for names. Does the latest commit seem okay?

retabak:
quick update: Working on a list of helper text! Trying to decide if we want placeholder text for all Transformers. Stay tuned...

retabak:
The BS:DS curriculum leans heavily on build, transform, filter, and sort - so it's no problem to come up with helper text for these. Here's what I've got (and these are also consistent with how naming is introduced in Pyret BS:DS):

Build Attribute: build-kilos
Transform Attribute: lbs-to-kg
Filter: filter-is-heavy
Sort: order-by-pounds

As for the many, many other Transformers - given that y'all are most familiar with how they work and what they do... would it be possible for ya'll to suggest text? If possible, those defaults could be drawn from / inspired by the [Animals dataset](https://docs.google.com/spreadsheets/d/1VeR2_bhpLvnRUZslmCAcSRKfZWs_5RNVujtZgEl6umA/edit#gid=0). Let me know what you think.

thomascastleman (approved):

PR #238, opened 2022-03-11
Support two kinds of sort

### LINKED ISSUE #224: Sort Transformer - description/UI
- [ ] What’s the difference between a “key expression” and a “formula expression”? If there is one, can it be explained here?
- [ ] What’s needed is not really an expression at all - it’s an attribute. 
- [ ] Also, the Domain of Sort ought to be “Table" -- but "row" is provided.

stardust66:
Resolves #224.

Changes:
- Add toggle component which is a dropdown where each option hides certain other components.
- Add sorting by attribute.

TODO:
- [x] Initialization issue: toggle value (by attribute or by expression) is initialized to null, even though it looks like it is set to "byAttribute" by default. If the user does not touch the toggle, sort will fail because toggle is null.
- [x] Comparison function issue: we need to be looser with types in our comparison function. Currently we do not detect numeric strings and sort them as strings instead.

### DIFF
--- a/src/components/transformer-template/TransformerTemplate.tsx
+++ b/src/components/transformer-template/TransformerTemplate.tsx
@@ -603,20 +603,22 @@ const TransformerTemplate = (props: TransformerTemplateProps): ReactElement => {
           return tmp ? (
             <div className="input-group">
               {titleFromComponent(component, init)}
-              <Select
+              <select
                 onChange={(e) => {
                   notifyStateIsDirty();
                   setState({ [component]: e.target.value });
                 }}
-                options={Object.entries(tmp.options).map(([key, values]) => ({
-                  title: values.title,
-                  value: key,
-                }))}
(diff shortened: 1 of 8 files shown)


stardust66:
Available for testing here: https://codap.concord.org/app/static/dg/en/cert/index.html?di=https://idyllic-cat-9dfb8b.netlify.app/

thomascastleman (approved):
Looks great!

retabak:
Just seeing this! Yay! Two types of Sort!
My one issue that when you click on "Sort method," whichever method is selected appears twice. 

Screenshot (1)

stardust66:
> My one issue that when you click on "Sort method," whichever method is selected appears twice.
Just fixed this! The preview should be updated as well.

retabak:
yay! Looks good to me.

P-bibs (approved):

PR #240, opened 2022-05-23
Allow new attribute name for transformed attribute

thomascastleman:
Users can now provide a new name for the attribute they are transforming with Transform Attribute, which makes sense because the transformed version may have a totally different interpretation or type from the original.

Transform-Attribute-New-Name


stardust66 (approved):
This looks good!

PR #241, opened 2022-05-23
Use consistent language for formula titles

thomascastleman:
Addresses the wording inconsistencies mentioned in #231 among other places.

Branched off of `sort` because the changes are dependent on the new Sort that has two variants.

### DIFF
--- a/src/transformerList.ts
+++ b/src/transformerList.ts
@@ -213,7 +213,7 @@ const transformerList: TransformerList = {
           title: "Dataset to Filter",
         },
         typeContract1: {
-          title: "How to Filter",
+          title: "Formula to Filter By",
           inputTypes: "Row",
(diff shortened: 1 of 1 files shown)


P-bibs (approved):
Looks good! I'm wondering if we might also want to change the messages for building/transforming attributes. Do we think either of these are noticeably better? If there isn't something that's a noticeable improvement then I don't think this is important enough to obsess over:
* Build Attribute: `Formula for Attribute Values` -> `Formula for New Attribute Values`
* Transform Attribute: `Formula for Transformed Values` -> `Formula for Transformed Attribute Values`

thomascastleman:
Those seem more consistent to me - I've made the changes

stardust66 (approved):

PR #245, opened 2022-06-01
Prevent saving transformers with duplicate names

P-bibs:
Adds an error message when trying to save a transformer with the same name as an existing transformer

image

### DIFF
--- a/src/components/transformer-template/DefinitionCreator.tsx
+++ b/src/components/transformer-template/DefinitionCreator.tsx
@@ -19,9 +19,11 @@ export default function DefinitionCreator({
   const [saveErr, setSaveErr] = useState<string | null>(null);
 
   async function saveTransformer(data: TransformerSaveData) {
-    const { name, purposeStatement } = data;
+    let { name, purposeStatement } = data;
 
-    if (name.trim() === "") {
+    name = name.trim();
+
+    if (name === "") {
       setSaveErr("Please give the transformer a name before saving.");
       return;
     }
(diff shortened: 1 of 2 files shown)


src/components/transformer-template/DefinitionCreator.tsx
```diff
@@ -29,139 +16,56 @@ export default function DefinitionCreator({
...
     } as SavedTransformerContent;
 
-    const savedTransformer = { name, description, content };
+    const savedTransformer = { name, purposeStatement, content };
     const encoded = encodeURIComponent(JSON.stringify(savedTransformer));
 
     const savedUrl = new URL(window.location.toString());
     savedUrl.searchParams.append("transform", encoded);
-
+    console.log(name);
```
thomascastleman, line 49: Maybe left by accident?

P-bibs:
Yep this is a good call I'll work on it. One other thing: should we trim all whitespace before saving the transformer's name? We do this to make sure you can't save a transformer with a name that's only whitespace but we don't actually trim whitespace when saving the transformer name.

thomascastleman:
Yeah, I was thinking about that. I think it could be good to both check against and save the trimmed name.

P-bibs:
@thomascastleman added trimming and duplicate suppression for editing saved transformers. Can you take a look when you get the chance?

thomascastleman (approved):
Looks good now.

One minor annoyance is that if you *edit* a saved Transformer's name to include leading or trailing whitespace, its name does change but if you click edit again the whitespace remains in the text input. It would be nice if the input contents matched the stripped contents, but I don't think this is a huge issue.

P-bibs:
Good catch, I believe that should be fixed now.

stardust66 (approved):
Looks good!

